PR #78, opened 2026-03-23
Verify daemon identity before session-cli restarts it

benvinegar:
## Summary
- persist a per-daemon identity record and expose an instance id in `/health`
- only restart the daemon when the health payload matches a trusted Hunk-managed daemon record
- keep compatibility with already-running legacy daemons by verifying the target pid still looks like a `hunk ... mcp serve` process on Linux

## Testing
- bun run typecheck
- bun run lint
- bun test test/daemon-record.test.ts test/mcp-server.test.ts test/session-commands.test.ts
- bun test test/session-cli.test.ts -t "list/get/context expose live Hunk sessions through the daemon"
- bun test test/mcp-e2e.test.ts -t "a live Hunk session auto-starts the daemon and renders CLI comments inline"

### DIFF
--- a/src/mcp/daemonRecord.ts
+++ b/src/mcp/daemonRecord.ts
@@ -1,0 +1,117 @@
+import fs from "node:fs";
+import { tmpdir } from "node:os";
+import { dirname, join } from "node:path";
+import { resolveHunkMcpConfig, type ResolvedHunkMcpConfig } from "./config";
+
+export interface HunkDaemonRecord {
+  pid: number;
+  startedAt: string;
+  instanceId: string;
+}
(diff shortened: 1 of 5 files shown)


benvinegar:
Closing this stale MCP-era branch. The code it touched (`src/mcp/*`) has since been removed/replaced by the session broker + `hunk-session` stack, and daemon hardening/restart behavior has moved through later PRs including #329, #332, #333, #409, and #424. If we still want stricter restart identity checks, it should be a fresh patch against `src/session-broker/*` / `src/session/*` rather than rebasing this branch.

PR #161, opened 2026-04-03
Support hunk-targeted live comments

benvinegar:
## Summary
- allow `hunk session comment add` to target a hunk directly with `--hunk <n>`
- resolve hunk-targeted comments to the first changed line on a stable side inside the live session
- update the repo skill and docs so agents know they can annotate whole hunks without hand-picking a line number

## Testing
- `bun run typecheck`
- `bun test`

Part of #158.

*This PR description was generated by Pi using OpenAI o3*

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ Use it to:
 - inspect the current review context
 - jump to a file, hunk, or line
 - reload the current window with a different `diff` or `show` command
-- add, list, and remove inline comments
+- add, list, and remove inline comments (by hunk or by line)
 
 Most users only need `hunk session ...`. Use `hunk mcp serve` only for manual startup or debugging of the local daemon.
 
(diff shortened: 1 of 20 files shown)


benvinegar:
Thanks for putting this together. I’m going to close this as superseded/stale rather than ask for another rebase, because the main agent-comment workflow from #158 has since landed through smaller follow-ups:

- #162 added bulk live-comment apply for agent reviews and closed #158.
- #281 added first-class session user review notes.
- Current main also supports hunk-level targets in the batch `hunk session comment apply` JSON shape.

This branch still contains an older MCP-era implementation and now conflicts with the session-broker/session-comment code that replaced it. The one piece that may still be useful is the direct CLI ergonomics for `hunk session comment add --hunk <n>`. If you still want that remaining functionality, please feel free to reopen or send a fresh narrow PR just for `comment add --hunk`, and we can review it against current main.

PR #203, opened 2026-04-14
fix: rebalance Pierre word highlights in split view

benvinegar:
## Summary

- keep Pierre word-level highlights visible in split view by deriving inline emphasis backgrounds from the diff row colors when theme defaults are too subtle
- stop the fallback blend at the first visible shade, with finer blend steps so the highlight stays noticeable without looking too bright
- add model- and render-level tests for inline word-diff emphasis and record the user-visible fix in `CHANGELOG.md`

## Testing

- bun run typecheck
- bun test

*This PR description was generated by Pi using OpenAI o4-mini*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


benvinegar:
@copilot resolve the merge conflicts in this pull request

PR #216, opened 2026-05-07
feat(ui): add pager sidebar toggle

clabby:
## Overview

Adds the ability to toggle the sidebar in pager mode. This makes navigating larger diffs a bit easier when using hunk as the pager for `jj` or `git`.


benvinegar:
Thanks @clabby!

PR #217, opened 2026-05-07
feat: Support `jj`

clabby:
## Overview

Adds support for [`jj`](https://docs.jj-vcs.dev/latest/) as an alternative VCS to `git`. The new `vcs` configuration option is optional, and falls back to `git` when unset for backwards compatibility with users' existing configuration files.

_This change was aided by Codex._


TGlide:
Amazing, would this work inside `jj` workspaces too?

clabby:
> Amazing, would this work inside `jj` workspaces too?

Yep!

benvinegar:
Thanks! Will be in next release

PR #222, opened 2026-05-07
fix(review): keep cross-file hunk navigation anchored

aliou:
Hello!

Noticed this bug a couple of days ago: when navigating between hunks with `[` or `]`, the pinned header could sometime revert to the previous/next file when close to the newly entered file.

- When going to the next hunk with `]`, and going to the next file, doing `]` again shows the previous file again
- When navigating back with `[`, it could jump prematurely to the previous file.

It's a bit hard to explain so I had gpt-5.5 make narrated repro videos:

<details><summary>Before</summary>
<p>

https://github.com/user-attachments/assets/d0879a3d-c4c5-4537-b0df-72fac7574579

</p>
</details> 

<details><summary>After</summary>
<p>

https://github.com/user-attachments/assets/07cf3924-e2d4-4423-aeee-f10c83f49451

</p>
</details> 

Let me know if you want / need other changes!

------

<details><summary>Summary by GPT-5.5:</summary>
<p>

## Summary

- Fix cross-file `]` hunk navigation so near-top hunks in the destination file cannot scroll above that file's body and briefly hand the pinned header back to the previous file.
- Fix cross-file `[` hunk navigation so backward jumps reveal the target hunk in the previous file instead of aligning that file's header to the top.
- Add regression coverage for the long-file/short-file navigation shape that reproduced the pinned-header flash and backward file-top jump.
- Document the user-visible fix in `CHANGELOG.md`.

## Files changed

- `src/ui/components/panes/DiffPane.tsx`: floors hunk and note reveal scroll targets to the selected file `bodyTop` before final viewport clamping.
- `src/ui/hooks/useReviewController.ts`: makes cross-file header alignment direction-aware, using file-top alignment only for forward jumps.
- `src/ui/AppHost.interactions.test.tsx`: adds forward and backward cross-file hunk navigation regression tests.
- `test/pty/harness.ts`: adds a PTY fixture with a long first file and short second file.
- `test/pty/ui-integration.test.ts`: adds live PTY regression coverage for backward cross-file hunk navigation.

## Verification

- `bun test src/ui/AppHost.interactions.test.tsx -t "cross-file hunk navigation"`
- `bun test test/pty/ui-integration.test.ts -t "backward cross-file hunk navigation reveals the target hunk in a real PTY"`
- `bun run typecheck`
- `bun run format:check`

Also verified the new tests fail when the fix files are stashed and pass after restoring the fix.

</p>
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Fixed
 
+- Fixed cross-file hunk navigation so near-boundary jumps keep the selected file pinned and backward jumps reveal the target hunk instead of the file top.
+
 ## [0.10.0] - 2026-04-21
 
 ### Added
--- a/src/ui/AppHost.interactions.test.tsx
+++ b/src/ui/AppHost.interactions.test.tsx
@@ -238,6 +238,51 @@ function createTwoFileHunkBootstrap(): AppBootstrap {
   });
 }
 
+/** Build the cross-file hunk-navigation shape that used to flash the previous pinned header. */
+function createCrossFileHunkNavigationBootstrap(): AppBootstrap {
+  const longBeforeLines = Array.from(
+    { length: 342 },
+    (_, index) => `line ${String(index + 1).padStart(3, "0")}`,
+  );
+  const longAfterLines = [...longBeforeLines];
+  for (const lineNumber of [
+    2, 21, 41, 61, 81, 101, 121, 141, 161, 181, 201, 221, 241, 261, 281, 301, 321, 341,
+  ]) {
+    longAfterLines[lineNumber - 1] = `line ${String(lineNumber).padStart(3, "0")} changed`;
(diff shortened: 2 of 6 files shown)


benvinegar:
@aliou Video was super helpful. What'd you use to make it?

aliou:
@benvinegar cheers! it's with gpt 5.5 + tmux + screen recording! I saw the bug a couple of weeks ago and forgot exactly what were the steps to repro, so when I resumed my initial research session, I asked the agent to use tmux to repro the bug so that i could attach to the session and confirm it was still visibly occurring  

then, I remembered that tmux can do popups, so I asked if it could use it to narrate what it's doing instead of relying on the comments in the example patch only.

Here's the full session, it's a bit all over the place and spans a couple of days, so linking directly to the start of the demo setup: https://pi.dev/session/#59e515c9749e735bbe7d8b9c4e276df4&leafId=17ca8ec3&targetId=94b15ab0

PR #225, opened 2026-05-07
docs: update agent-workflows.md according to latest version

smoya:
Since `0.10.0` is out, `hunk skill path` cmd is available. This PR updates the docs to encourage the user to use it.


benvinegar:
Thanks!

PR #228, opened 2026-05-07
fix(loaders): strip git log -p commit metadata before parsing

gonzaloserrano:
Piping `git log -p` (or `git show -p` of multiple commits) through hunk's pager mode floods stderr with hundreds of
`parseLineType: Invalid firstChar` / `processFile: invalid rawLine` warnings from @pierre/diffs.

Root cause: pager mode's `looksLikePatchInput` heuristic accepts the stream because it contains `diff --git` and `@@`, then forwards the whole blob to `parsePatchFiles`. But @pierre/diffs is a strict patch parser: every hunk-body line must start with `+`, `-`, ` ` or `\`. The `commit <sha>`, `Author:`, `Date:`, blank, and 4-space-indented commit message lines that appear between commits all fail that check, and the parser logs each rejection unconditionally (no quiet flag).

Fix in `normalizePatchChangeset` rather than in `pager.ts` so all input paths benefit (a saved `git log -p` fed via `hunk patch file.txt` would have hit the same bug). The new `stripGitLogMetadata`:

- Fast-paths to a no-op when no `^commit [0-9a-f]+` boundary exists, keeping the regular patch path zero-allocation.
- State-machines per line: enters "header" mode on a commit boundary, drops every following line until a patch start (`diff --git `, `--- `, `+++ `).
- Preserves context lines that mention "commit" textually — they start with the diff line-type marker (` `/`+`/`-`), so the `^commit ` regex doesn't match.

Tests cover: single-commit, multi-commit, decorated headers (refs in parens), abbreviated SHAs, --stat blocks, merge metadata, context lines mentioning "commit", trailing diff-less commit, and an end-to-end check that the stripped output round-trips through `parsePatchFiles` with zero `Invalid firstChar` warnings.


benvinegar:
Thanks!

PR #229, opened 2026-05-07
docs(agent): note sandboxed session access

skalidindi:
Adds a short troubleshooting note for agent workflows where `hunk session *` cannot reach the local loopback daemon from inside a sandbox.

  I hit this specifically with Codex sandbox mode: `hunk session list` reported no active sessions even though a Hunk TUI was running and the daemon's `/session-api` endpoint showed a
  registered session. This can make the normal "No active Hunk sessions" guidance misleading.

  The docs now suggest probing `/session-api` directly and rerunning session commands with the agent's local-network or sandbox-escalation mechanism when the daemon does have registered
  sessions.

  I’d avoid mentioning Claude unless someone confirms it has the same behavior.

### DIFF
--- a/docs/agent-workflows.md
+++ b/docs/agent-workflows.md
@@ -25,6 +25,18 @@ When a Hunk TUI starts, it registers with a local loopback daemon. `hunk session
 
 Most users only need `hunk session ...`. Use `hunk mcp serve` only for manual startup or debugging of the local daemon.
 
+Some agent sandboxes restrict localhost access. If an agent reports no active sessions while a
+Hunk TUI is visibly running, first check whether the agent can reach the loopback daemon:
+
+```bash
+curl -s -X POST http://127.0.0.1:47657/session-api \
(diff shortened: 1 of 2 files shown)


benvinegar:
I feel like I just want to abandon the HTTP server and move to a socket instead. Would that solve the sandbox issue?

skalidindi:
Yes, I think that is the cleaner direction and should avoid the specific Codex failure mode because the agent would no longer need loopback TCP access. Note that the Unix-domain socket would also still need to live somewhere visible to both processes like the existing runtime dir and have sensible per-user permissions for the agent to see it in sandbox mode.

If you want to replace the transport soon, I’m happy to either close this or keep it as a temporary solution until then. Fine either way.

benvinegar:
Well, the HTTP server is because this originally was an MCP server. I might wanna keep that around. I think in the meantime I'll maybe tweak your file to have lighter guidance around debugging (reduce tokens).

PR #230, opened 2026-05-07
feat(nix): add flake to support nix builds

### LINKED ISSUE #215: Would you be able to add a Nix Flake?
Hey, would you be open to adding a flake.nix so nix users don't have to package it themselves? I can contribute it. Thanks for your time!

emrtnn:
Closes #215 

As suggested in the issue, this PR adds a flake.nix file that Nix users can use to natively install and run the package.

The way I structured it is a top-level `flake.nix` that imports `.nix/package.nix`, the file containing the actual logic for building Hunk.

This is designed so no core maintainer needs any Nix knowledge:
Inside `.nix/`, there is a `nix-deps-hash.txt` file that's generated by the new `bun run nix:update-hash` script. Whenever the project's dependencies change, the Nix build will temporarily fall out of sync until someone runs this script to fetch the new hash. I suggest automating this via CI in the future so it's entirely hands-off for you, but I didn't want to make this initial PR too broad.

Since @MarkusZoppelt has already opened a PR (NixOS/nixpkgs#517626) to officially add Hunk to nixpkgs, this flake can be used by the community in the present while we wait for that to merge. Later, it will remain highly useful for local development (nix develop) and building purposes directly from source (the local building strategy is different from how nixpkgs will fetch the binary). I also used the exact same package name he chose, as discussed in his PR.

### DIFF
--- a/.nix/nix-deps-hash.txt
+++ b/.nix/nix-deps-hash.txt
@@ -1,0 +1,1 @@
+sha256-Lr3M22JlMcX+HTZay2A1rN/zwGnznIjaib3E3y0Aob0=
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,35 @@
+{
+  description = "Hunk - Review-first terminal diff viewer for agentic coders";
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+    systems.url = "github:nix-systems/default";
+  };
+
+  outputs = {
+    systems,
+    nixpkgs,
+    ...
+  }: let
+    eachSystem = nixpkgs.lib.genAttrs (import systems);
+  in {
+    packages = eachSystem (
+      system: let
+        pkgs = import nixpkgs {inherit system;};
+      in {
+        default = import ./.nix/package.nix {inherit pkgs;};
+      }
+    );
+    devShells = eachSystem (
+      system: let
+        pkgs = import nixpkgs {inherit system;};
+      in {
+        default = pkgs.mkShell {
+          buildInputs = with pkgs; [
+            bun
+            nodejs
+          ];
+        };
+      }
+    );
+  };
+}
--- a/flake.nix
+++ b/flake.nix
@@ -27,6 +27,7 @@
           buildInputs = with pkgs; [
             bun
             nodejs
+            git
           ];
         };
       }
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ node_modules
 out
 dist
 *.tgz
+result
 
 # code coverage
 coverage
--- a/.nix/package.nix
+++ b/.nix/package.nix
@@ -1,0 +1,65 @@
+{pkgs}: let
+  packageData = builtins.fromJSON (builtins.readFile ../package.json);
+
+  # Fetch dependencies (Network enabled, verified by hash)
+  bunDeps = pkgs.stdenv.mkDerivation {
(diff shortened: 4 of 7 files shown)


m4r1vs:
Hey, I just opened basically the same PR, only using bun2nix instead of your impure network-enabled solution. I've had very good experience with bun2nix in the past. I think it would be a little annoying having to update the hash every time a package is updated -- especially by people who don't use nix.

I can see that having the bun.lock.nix file which bun2nix requires could also be confusing for people not using nix. Also the added `postinstall` script. 

If this PR is the way forward, I'd suggest adding your bash script to `postinstall` such that the bunDeps hash is automatically updated. Otherwise it looks good to me.

flake.nix
```diff
@@ -0,0 +1,35 @@
...
+      }
+    );
+    devShells = eachSystem (
+      system: let
+        pkgs = import nixpkgs {inherit system;};
+      in {
+        default = pkgs.mkShell {
+          buildInputs = with pkgs; [
+            bun
+            nodejs
```
m4r1vs, line 29: Also nodejs is not needed in the devShell I think -- even though it's listed as a requirement in CONTRIBUTING.md. Not sure why it is. Git should always be installed but could be added to the devShell as to be explicit.

emrtnn: Hey, thanks for the comment! Yeah indeed I tried yesterday to build it using bun2nix but adding the postinstall script gave me a lot of issues because this repo uses bun as a dependency itself so it was causing a conflict between the nix installed bun and the repo installed bun by the nix one. Since I couldn't manage solve this I chose this way.

image

Didn't know you were also working on this if I had I wouldn't have made this PR, that's why I opened #215 before anything.

That said, if you agree with following with this PR I consider your `postinstall` suggestion a really good catch and will be happy to add it. The only issue I see is that it will make installing deps much slower but maybe that's not that big of a deal for core contributors

m4r1vs: I think it's very cool how we took two different routes and both landed at a valid solution to the problem! I too ran into the issue with the `bun` dependency in `package.json`. However, I just removed it there xd

.nix/nix-deps-hash.txt
```diff
@@ -0,0 +1 @@
+sha256-Lr3M22JlMcX+HTZay2A1rN/zwGnznIjaib3E3y0Aob0=
```
m4r1vs, line 1: I think the hash might change across architectures (darwin/linux and x86/arm64). I can test this if needed

emrtnn: Removing it was also my first thought but I want to believe is there for an _unknown_ reason to us and I usually try touching core files like `package.json` as little as possible.

Now that you speak about that I think you are right too. I can also test tomorrow at work where my machine has an ARM CPU. If our concerns are right then I don't think this is anymore a viable method. We would have to add a `json` file with one hash per architecture instead of a `txt` one and that is a maintenance nightmare.

We should use bun2nix if that happens but I just want to be sure that there's no issue with deleting bun as a dependency in package.json so this PR can be closed in favor of yours.

Any core maintainer can give opinion on removing it?

m4r1vs: just tried it on arm64 fedora and indeed the hash is different :/

emrtnn:
Closing this PR after the node_modules hash discovery. As @elucid said we should focus on #231 and find a way to workaround the node_module's bun issue without removing it. Thanks to all for your time and feedback!

PR #231, opened 2026-05-07
feat(nix): Add development shell and package export

### LINKED ISSUE #215: Would you be able to add a Nix Flake?

m4r1vs:
I use nix and try to avoid globally installing npm packages.
There is a [pull request to add Hunk to nixpkgs upstream](https://github.com/NixOS/nixpkgs/pull/517626), ~~however it is stalled by support for direct bun packaging~~.

We use [bun2nix](https://github.com/nix-community/bun2nix) which I have had good experience with in the past.

Closes #215. 

I've tested this on x86 NixOS, arm64 Fedora with Nix installed and MacOS Tahoe with nix-darwin. ~~I ran into a couple of issues which I solved by removing `bun` as a dependency in `package.json`. Not sure why it's there -- I've never seen it installed like this before.~~

I tried to keep the documentation in README and CONTRIBUTING minimal and elaborated more in a new README in the `nix` directory.

Happy to answer any questions and to maintain the nix flake if maintanance is needed. Cheers :)

Ps. I also added a home-manager module so Hunk can be configured using nix. It also allows setting it as the default git pager so `git diff` works when using hunk through home-manager without any extra configuration needed.


package.json
```diff
@@ -69,11 +69,11 @@
     "bench:bootstrap-load": "bun run benchmarks/bootstrap-load.ts",
     "bench:highlight-prefetch": "bun run benchmarks/highlight-prefetch.ts",
     "bench:large-stream": "bun run benchmarks/large-stream.ts",
-    "bench:large-stream-profile": "bun run benchmarks/large-stream-profile.ts"
+    "bench:large-stream-profile": "bun run benchmarks/large-stream-profile.ts",
+    "postinstall": "bun2nix -o nix/bun.lock.nix -c ../"
```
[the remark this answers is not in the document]
m4r1vs: No, `-c` is relative to the parent dir of the `bun.nix` file so `../` is project root in this case. `nix build` fails otherwise. Also, it is intended that every user (using nix or not) updates the bun2nix lock. Otherwise we could potentially land in a state where bun2nix and package.json are not the same.

nix/devShell.nix
```diff
@@ -0,0 +1,7 @@
+{pkgs}:
+pkgs.mkShell {
+  buildInputs = with pkgs; [
+    bun
+    git
```
m4r1vs, line 5: nodejs is mentioned as a requirement in CONTRIBUTING.md. However, I'm not sure why it is. I've not added it yet because everything worked without nodejs for me.

package.json
```diff
@@ -69,11 +69,11 @@
...
   },
   "dependencies": {
     "@pierre/diffs": "^1.1.19",
-    "bun": "^1.3.10",
```
m4r1vs, line 77: Why was bun here in the first place?

elucid:
Hey @m4r1vs (cc @emrtnn). Thanks for the PRs and for the discussion. I've looked at both this and #230. #230's approach is simpler but as you've both discovered, the fixed node_modules hash is platform dependent. I think this PR's approach is more nix-native and more maintainable.

Having said that, you'll need to make some fixes before it can be merged:

1. don't remove bun. hunk uses bun throughout the source/build/dev workflow. a number of things would have to change if it was removed. we can debate about whether the project should use bun, but not within a PR to add nix support
2. make the nix build use nix's bun e.g. `${bun}/bin/bun build --compile ...`
3. the postinstall hook is too invasive, remove it.  we don't need to regenerate nix files on every install. add a `"nix:update-lock"` script instead.
4. don't run unnecessary lifecycle hooks (prepare / simple-git-hooks) during nix builds
5. fix the flake dev shell shape. it should expose `devShells.${system}.default` not a derivation directly at `devShells.${system}`
6. add package metadata similar to what #230 has
7. the nix package build and basic bun workflows don't need nodejs. but if this PR exposes `nix develop` as a contributor shell, it should match `CONTRIBUTING.md` and include nodejs too. otherwise, omit the devShell.

m4r1vs:
Thanks for the feedback! Appreciate it!
1. When bun is installed as a nodejs dependency, bun2nix does not work. I'm happy to open an issue at their repository but I also feel weird about having multiple different bun binaries potentially in conflict (one in node_modules, one from nix and then one often globally installed -- be it through homebrew, mise or bun's install script). Wouldn't it make sense to remove bun as a nodejs dependency and just list it instead of/together with nodejs as a requirement? I mean, we use bun.lock instead of package.lock. If we used the latter, we could make everything much simpler by using fetchNpmDeps directly from nixpkgs.
2. It should already default to the bun binary in nixpkgs within the nix builder, but I'll check.
3. Will remove that, 👍 Removing this could also fix 1) because bun2nix was failing as a postinstall script -- not within the build itself.
4. Yes, thanks
5. 👍
6. 👍

m4r1vs:
regarding 3) I think it does not run them by default (see [here](https://github.com/nix-community/bun2nix/blob/c843f477b15f51151f8c6bcc886954699440a6e1/docs/src/building-packages/hook.md?plain=1#L72)) but I'll make it explicit

m4r1vs:
@elucid I think I've been able to address all your points:
 - Added bun to deps, removed bun2nix as dependency and from postinstall. Now it can be run using `bun run nix:update-lock`. [commit](https://github.com/modem-dev/hunk/pull/231/changes/1a34d29e3be33178bd0a26b3db28c7323282f7b4)
 - Added CI/CD check to make sure nix lock is up to date with helpful error (automation probably overkill?) [commit](https://github.com/modem-dev/hunk/pull/231/changes/d6202afe370aa307897dfbc2236d3005efed9e00)
 - Added nodejs to devShell
 - Fixed devShell shape
 - Made skipping lifecycle hooks explicit
 - Added package metadata (copied from #230 ) [commit](https://github.com/modem-dev/hunk/pull/231/changes/3580f9693c771b75f8856d6b9012ba253b7329a7)

Cheers!

I think the removal of postinstall bun2nix should also fix the [failed pipeline](https://github.com/modem-dev/hunk/actions/runs/25527148091/job/74942673777)

m4r1vs:
I've noticed that `hunk skill path` is not working when installed through nix. It walks up from the Hunk binary location so  I fixed it by copying SKILLS to $out.

A more idiomatic way would be to copy it to $out/share but that would mean, we'd have to modify `src/core/paths.ts` to look for the Skills in that directory possible breaking non-nix installs.

PR #232, opened 2026-05-07
docs(readme): add keyboard shortcuts guide

hugows:
Had a hard time finding out keyboard shortcuts, for some reason.


benvinegar:
I agree with mentioning the shortcuts, but I don't think this is how I would place them in the readme. I'm particular over how the README looks. I'll probably take this and tune it.

benvinegar:
There's now a docs page with this: https://www.hunk.dev/docs/start/keyboard-and-mouse/

PR #236, opened 2026-05-07
fix(ui): align sidebar menu state with visibility

aliou:
Hello!

Noticed this UI bug while interacting with Hunk from a smaller terminal window:

- by default, the sidebar display state is set to displayed. this is what is used to show the checkbox in the View menu
- however, rendering also checks if the window can display the sidebar based on the width. if not, it doesn't render it
- the issue is that the menu item still says the sidebar is displayed even if it is not actually rendered

This PR makes the checkbox reflect whether the sidebar is actually rendered.

Research + code review session (gpt-5.5): https://pi.dev/session/#bcf0b5045252c82447828797ca8e013c 
Implementation session (kimi k2.6): https://pi.dev/session/#758e5740c6f72a5289d28f0643937396
-> Some manual code tweaks were done outside of those two sessions (especially around tests)

Let me know if you need/want more changes!

------

<details><summary>Summary by GPT-5.5:</summary>
<p>

## Summary

- Initialize sidebar visibility from the resolved responsive layout in `src/ui/App.tsx` instead of always defaulting to visible.
- Pass the effective `renderSidebar` state into `buildAppMenus` so the View menu checkmark reflects the sidebar currently rendered on screen.
- Update `src/ui/lib/appMenus.ts` and `src/ui/lib/ui-lib.test.ts` to use `renderSidebar` as the menu state input.
- Add AppHost responsive coverage in `src/ui/AppHost.responsive.test.tsx` for medium-width terminals where the sidebar is hidden by layout but can still be forced open.

## Behavior

On medium-width terminals, the View menu now shows Sidebar unchecked when the sidebar is not visible. Pressing `s` still opens the hidden sidebar when there is enough room to force it open.

## Validation

- `bun test src/ui/AppHost.responsive.test.tsx`
- `bun test src/ui/lib/ui-lib.test.ts`
- `bun run typecheck`

</p>
</details>


benvinegar:
Thanks @aliou!

PR #240, opened 2026-05-07
fix(loaders): restore a/b prefixes on noprefix patch input

mo:
## Summary

`hunk pager` and `hunk patch` show "No files match the current filter." for any input produced by a `git diff` that ran with `diff.noprefix=true`. The diff itself reaches Hunk fine — the parser just rejects it.

Repro (in any repo, in a TTY):

```bash
git config --global diff.noprefix true
git config --global core.pager 'hunk pager'
git diff HEAD~1..HEAD     # → "No files match the current filter."
```

Same symptom via the patch path:

```bash
git -c diff.noprefix=true diff HEAD~1 | hunk patch -
```

## Root cause

`@pierre/diffs` `processFile()` runs:

```js
if (line.startsWith("diff --git")) {
  const [, , prevName, , name] = line.trim().match(ALTERNATE_FILE_NAMES_GIT) ?? [];
  currentFile.name = name.trim();   // TypeError when match is null
  ...
}
```

`ALTERNATE_FILE_NAMES_GIT` requires literal `a/` and `b/` prefixes:

```
/^diff --git (?:"a\/(.+?)"|a\/(.+?)) (?:"b\/(.+?)"|b\/(.+?))$/
```

With `diff.noprefix=true`, git emits `diff --git src/x.ts src/x.ts` (no `a/`/`b/`). The regex fails, `name` is `undefined`, `.trim()` throws `TypeError`, and `parsePatchFiles` rethrows because Hunk passes `throwOnError: true`. The outer `try/catch` in `normalizePatchChangeset` swallows the throw and returns `files: []` — hence the empty review.

## Why it only affects the patch path

`src/core/git.ts` already defends git-backed reviews with `DIFF_PREFIX_NORMALIZATION_ARGS`:

```ts
const DIFF_PREFIX_NORMALIZATION_ARGS = [
  "-c", "diff.noprefix=false",
  "-c", "diff.mnemonicPrefix=false",
  "-c", "diff.srcPrefix=a/",
  "-c", "diff.dstPrefix=b/",
];
```

That covers `hunk diff`, `hunk show`, `hunk stash show` because Hunk runs git itself. It can't cover the patch path, where the patch text was already produced by an outer git process (the user's `core.pager = hunk pager`, or a literal `git diff | hunk patch`).

There is also a hint of this in the existing test suite: `loads colorized git patch files like the real pager stdin stream` shells out to `git diff --no-index --color=always` and feeds the result to `loadAppBootstrap`. On any machine with `diff.noprefix=true` set globally, that test fails today (`Expected length: 1, Received length: 0`).

## Fix

Add `normalizeGitPatchPrefixes()` to `src/core/loaders.ts` and run it inside `normalizePatchChangeset` before `parsePatchFiles`. The helper:

- Walks the patch line by line, scoped per `diff --git` block.
- Detects `diff --git` headers in noprefix form (quoted or unquoted, single- or multi-token paths) and rewrites them to canonical `a/X b/Y`.
- Tracks whether the current block actually needed rewriting and only rewrites the corresponding `---`/`+++` lines in that block. That keeps it safe for paths that legitimately live inside an `a/`-named directory and were already prefixed (`diff --git a/a/inner.ts b/a/inner.ts`).
- Leaves `/dev/null` untouched on the unified-diff lines.
- Leaves the one genuinely ambiguous case alone: a rename of unquoted paths that both contain spaces (`diff --git old name.txt new name.txt`). Git itself can't unambiguously round-trip that with `diff.noprefix=true`, so we don't try to.

## Tests

Three new tests in `src/core/loaders.test.ts` exercise the patch path directly via `kind: "patch"`:

- noprefix non-rename input parses to one file.
- noprefix rename input parses to one renamed file with `previousPath` recovered.
- already-prefixed input where the path lives in an `a/`-named directory is preserved unchanged.

The previously-failing `loads colorized git patch files like the real pager stdin stream` test passes on machines that have `diff.noprefix=true` set globally.

## Test plan

- [x] `bun run typecheck`
- [x] `bun test src/core/` — 140 pass, 0 fail
- [x] `bun test src/core/loaders.test.ts` — 37 pass, 0 fail (3 new)
- [x] `bun run lint` — 0 warnings, 0 errors
- [x] `bun run format`
- [x] Manual: `git config --global diff.noprefix true && git diff HEAD~3..HEAD | hunk pager` now renders the diff instead of the empty-state message.
- [ ] CI green

### DIFF
--- a/src/core/loaders.test.ts
+++ b/src/core/loaders.test.ts
@@ -1037,6 +1037,33 @@ describe("loadAppBootstrap", () => {
     });
   });
 
+  test("loads quoted noprefix patch text emitted for escaped git paths", async () => {
+    const dir = createTempRepo("hunk-patch-quoted-noprefix-");
+    const fileName = "src\tfile.txt";
+
+    writeFileSync(join(dir, fileName), "one\n");
+    git(dir, "add", ".");
+    git(dir, "commit", "-m", "initial");
+
+    writeFileSync(join(dir, fileName), "two\n");
+    const patchText = git(dir, "-c", "diff.noprefix=true", "diff", "--", fileName);
+
+    expect(patchText).toContain('diff --git "src\\tfile.txt" "src\\tfile.txt"');
+
+    const bootstrap = await loadAppBootstrap({
+      kind: "patch",
+      text: patchText,
+      options: { mode: "auto" },
+    });
+
+    expect(bootstrap.changeset.files).toHaveLength(1);
+    expect(bootstrap.changeset.files[0]).toMatchObject({
+      path: "src\\tfile.txt",
+      metadata: { name: "src\\tfile.txt", type: "change" },
+    });
+    expect(bootstrap.changeset.files[0]?.stats).toEqual({ additions: 1, deletions: 1 });
+  });
+
   test("does not mangle a deleted SQL `-- comment` line in a noprefix patch", async () => {
     // The original source line `-- drop table users;` (a SQL comment) is encoded in a unified
     // diff deletion as `--- drop table users;` — three dashes (one for the deletion marker,
--- a/src/core/loaders.test.ts
+++ b/src/core/loaders.test.ts
@@ -993,4 +993,66 @@ describe("loadAppBootstrap", () => {
     expect(bootstrap.changeset.files[0]?.path.endsWith("after.ts")).toBe(true);
     expect(bootstrap.changeset.files[0]?.stats.additions).toBeGreaterThan(0);
   });
+
+  test("loads patch text emitted with diff.noprefix=true (e.g. from `hunk pager` stdin)", async () => {
+    const bootstrap = await loadAppBootstrap({
+      kind: "patch",
+      text: [
+        "diff --git src/example.ts src/example.ts",
+        "index 0000000..1111111 100644",
+        "--- src/example.ts",
+        "+++ src/example.ts",
+        "@@ -1,1 +1,2 @@",
+        " const value = 1;",
+        "+const added = 2;",
+      ].join("\n"),
+      options: { mode: "auto" },
+    });
+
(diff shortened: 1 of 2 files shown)


mo:
Thanks @greptile-apps — confirmed real bug, fixed in bf21291.

For anyone tracking, the corruption I could reproduce is slightly different from "the file disappears": with the original commit a noprefix patch that deletes a SQL `-- comment` line still rendered the file, but the deletion content showed as `-- a/drop table users;` instead of `-- drop table users;`. The line counts still added up so the parser didn't reject the patch — it just silently displayed the wrong text on the left side of the diff. Either way, the suggested fix (clearing `blockNeedsPrefix` after `+++`) is the right call, and I've added a regression test that reproduces the corruption against the previous commit.

mo:
@benvinegar let me know if you have any questions about this PR, I'd be happy to explain the problem it fixes more in depth if necessary

benvinegar:
@mo Sorry just trying to catch up to everything :) Looking ...

benvinegar:
Added a follow-up commit (`fa323e9`) for quoted noprefix paths. Git quotes escaped paths like `src\tfile.txt`, and Pierre still rejects quoted `diff --git` headers, so the normalizer now canonicalizes those headers to its unquoted `a/... b/...` form. I also added a real-git regression test for `git -c diff.noprefix=true diff` with a tabbed filename.

Verified with `bun test src/core/loaders.test.ts`, `bun run typecheck`, and `bun run lint`.

benvinegar:
Rebased this PR onto current `origin/main` and resolved the loader conflict by applying prefix normalization after the newer git-log metadata stripping path.

Verified after the rebase with `bun test src/core/loaders.test.ts`, `bun run typecheck`, and `bun run lint`.

benvinegar:
Thanks @mo!

PR #241, opened 2026-05-07
feat(ui): add new theme "auto" that automatically switches between da…

### LINKED ISSUE #238: Automatic light/dark theme switching
All the (terminal) programs I use support automatic switching between dark and light theme based on system preferences or terminal background. I think OpenTUI apps can support this too, it would be cool if Hunk could do this.

m4r1vs:
fixes #238

- Introduced a new `auto` theme
- Added a "Follow system" toggle in the UI theme menu.
- Implemented a theme resolution strategy that selects between a designated `theme_light` and `theme_dark` based on the detected terminal or system mode (with support for pipe-mode, e.g. when `git diff | hunk patch -` is used.
- Tested on MacOS and Linux Desktop

image
image

### Technical details
- Added `detectSystemThemeMode` to fall back to OS-level settings (starting with macOS support via `defaults read -g AppleInterfaceStyle`) when terminal-based detection is inconclusive (e.g., in piped/pager flows).
- Integrated `ThemeMode` listeners in the React `App` shell to respond to live system theme changes.
- Updated `README.md` and component documentation to reflect the new theme options.

### DIFF
--- a/src/core/terminal.ts
+++ b/src/core/terminal.ts
@@ -39,6 +39,26 @@ export function shouldUseMouseForApp({
   return stdinIsTTY || hasControllingTerminal;
 }
 
+/** Detect the system theme mode as a fallback when terminal-based detection is unavailable. */
+export function detectSystemThemeMode(
+  deps: { spawnSync: typeof Bun.spawnSync } = { spawnSync: Bun.spawnSync },
+): "light" | "dark" | undefined {
+  if (process.platform !== "darwin") {
+    return undefined;
+  }
+
+  try {
+    // defaults read -g AppleInterfaceStyle returns "Dark" in dark mode.
+    // It exits with code 1 if the key doesn't exist (which means light mode).
+    const proc = deps.spawnSync(["defaults", "read", "-g", "AppleInterfaceStyle"]);
+    const output = proc.stdout.toString().trim();
+    return output === "Dark" ? "dark" : "light";
+  } catch {
+    // On macOS, if the key is missing, it's light mode.
+    return "light";
+  }
+}
+
 export interface ControllingTerminal {
   stdin: tty.ReadStream;
   close: () => void;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,14 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Added
 
+- Added automatic switching between dark and light mode themes based on system preferences.
+- Added a "Follow system" option to the theme menu.
+
 ### Changed
 
 ### Fixed
+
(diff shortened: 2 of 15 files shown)


src/core/terminal.ts
```diff
@@ -39,6 +39,26 @@ export function shouldUseMouseForApp({
   return stdinIsTTY || hasControllingTerminal;
 }
 
+/** Detect the system theme mode as a fallback when terminal-based detection is unavailable. */
+export function detectSystemThemeMode(
+  deps: { spawnSync: typeof Bun.spawnSync } = { spawnSync: Bun.spawnSync },
+): "light" | "dark" | undefined {
+  if (process.platform !== "darwin") {
+    return undefined;
```
m4r1vs, line 47: I don't know why the pager/piped flow works on Linux but it somehow does lol

m4r1vs:
Ran into an issue where it would use the dark theme when opening it immediately after switching from dark to light in system preferences on MacOS. I'll undraft when I've fixed it.

src/core/terminal.ts
```diff
@@ -39,6 +39,26 @@ export function shouldUseMouseForApp({
...
+  }
+
+  try {
+    // defaults read -g AppleInterfaceStyle returns "Dark" in dark mode.
+    // It exits with code 1 if the key doesn't exist (which means light mode).
+    const proc = deps.spawnSync(["defaults", "read", "-g", "AppleInterfaceStyle"]);
```
dive, line 53: I find this idea a bit tricky. Some people use macOS in a light theme but still want a dark theme for the terminal. To me, the terminal should be the main source of information to keep things clear. For instance, you could use [CSI](https://contour-terminal.org/vt-extensions/colour-palette-update-notifications/) to notify the tool when the colour palette changes. OpenTUI support it via the [theme mode](https://opentui.com/docs/core-concepts/renderer/).

m4r1vs: yeah, but then you could just not use the auto theme, no?

The issue with the OpenTUI theme mode is that it does not work when Hunk is used as a pager because stdin is redirected

dive: Fair enough. It is just odd that Hunk will not follow the terminal when I change the light/dark themes outside of the OS general settings.

elucid:
Closing in favour of #290 which takes a different detection approach.

Instead of reusing OpenTUI's theme_mode signal and adding OS-level fallbacks around it, #290 probes the terminal background directly via OSC 11 using the same /dev/tty input stream that #198 established for mouse support. This avoids the stdin/mouse/theme coupling that caused the earlier regressions in piped mode.

PR #243, opened 2026-05-08
fix(ui): use full agent-note set for section geometry measurement

### LINKED ISSUE #234: Bottom-edge scroll snap-back when agent notes are attached
## Summary
With agent notes attached to a session, scrolling to the very bottom of the diff snaps back upward by ~the height of the note rows on files that just left the top of the viewport. Without notes, the bottom edge is stable. Repros in vanilla 0.10.0.

## Repro
1. `hunk diff <range>` on a multi-file diff
2. Attach agent notes to one or more files (`hunk session comment apply ...`)
3. Scroll all the way to the last line of the last file
4. Try to scroll one line further → viewport snaps up by 1–2 lines (matches the note-row count of files that just scrolled off the top)

Without step 2 the bottom edge is stable.

## Likely cause (read-only inspection of `src/ui/components/panes/DiffPane.tsx`)
- `sectionGeometry` (line 439) measures with `visibleAgentNotesByFile` and falls back to notes-less `baseSectionGeometry` for files outside the viewport.
- `visibleAgentNotesByFile` (line 415) is keyed off `visibleViewportFileIds`, which depends on `scrollViewport.top`.
- `totalContentHeight` (line 478) is derived from this geometry, so it shrinks by the height of off-top note rows as you scroll past them.
- `clampReviewScrollTop` (line 481) uses `totalContentHeight`. As content height shrinks, `maxScrollTop` shrinks, and any settle / clamp pass yanks the viewport upward.

The "off-screen optimization" that excludes notes from measurement is correct for *rendering* (don't paint hidden cards) but wrong for *measurement* — total content height needs to be stable regardless of scroll position. Files with notes attached should always have their note rows included in `sectionGeometry`, even when off-screen.

## Suggested fix
Use `allAgentNotesByFile` (not `visibleAgentNotesByFile`) for measurement at line 442 — keep `visibleAgentNotesByFile` for the render call at line 1115.

## Related
Different bug from #233 (max update depth), but same file. Both surface around viewport / scroll bookkeeping under agent annotations.

aldevv:
**User-visible:** with agent notes attached, scrolling to the last line of the diff bounces the viewport upward by ~1–2 lines (~the height of the off-top note rows).

**Steps to reproduce (on `main`)**
1. `hunk diff <range>` on a multi-file diff.
2. Attach agent comments on at least one file other than the last one (so notes scroll off the top while you reach the bottom):
   ```sh
   echo '{"comments":[{"filePath":"<early-file>","hunkNumber":1,"summary":"x","rationale":"y"}]}' \
     | hunk session comment apply --repo <repo> --stdin
   ```
3. Scroll all the way down to the last line of the last file.
4. Try to scroll one line further. The viewport snaps up by ~1–2 lines instead of staying put.

Without notes attached, the bottom edge is stable.

**Fix:** section geometry now measures with the full agent-note set so total content height is stable regardless of scroll position. Rendering still uses the viewport-restricted set, so painting cost is unchanged.

No user-facing docs or workflows change.

Fixes #234.

### DIFF
--- a/src/ui/components/panes/DiffPane.tsx
+++ b/src/ui/components/panes/DiffPane.tsx
@@ -415,11 +415,18 @@ export function DiffPane({
     return next;
   }, [allAgentNotesByFile, selectedFileId, showAgentNotes, visibleViewportFileIds]);
 
+  // Measure with the *full* set of agent notes per file, not just the visible-viewport set.
+  // The visible set is correct for rendering (skip painting cards on off-screen files), but
+  // using it here makes total content height fluctuate with scroll position: as a file with
+  // notes leaves the viewport, its measurement shrinks back to the no-notes baseline, which
+  // shrinks `totalContentHeight`, which tightens `clampReviewScrollTop`'s ceiling, which
+  // snaps the viewport upward by the height of the off-top note rows. Always include notes
+  // in geometry for stable bottom-edge clamping.
   const sectionGeometry = useMemo(
     () =>
       files.map((file, index) => {
-        const visibleNotes = visibleAgentNotesByFile.get(file.id) ?? EMPTY_VISIBLE_AGENT_NOTES;
-        if (visibleNotes.length === 0) {
+        const notes = allAgentNotesByFile.get(file.id) ?? EMPTY_VISIBLE_AGENT_NOTES;
+        if (notes.length === 0) {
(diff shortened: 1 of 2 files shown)


benvinegar:
@aldevv Can you submit a test case?

PR #251, opened 2026-05-08
fix(ui): stabilize agent note navigation

### LINKED ISSUE #233: Maximum update depth exceeded under rapid viewport changes (scroll or hunk-jump)
## Summary
Hunk 0.10.0 crashes with "Maximum update depth exceeded" when the viewport is updated rapidly. Reproduces on both `mode = "auto"` and `mode = "stack"`; triggered by either scrolling or hunk-jumping (next/prev hunk navigation).

## Version
0.10.0

## Repro (either path)
- `hunk diff <range>` on a diff with several hunks across multiple files
- Either: scroll continuously, OR press the next/prev-hunk key several times in quick succession
- Crash fires within a handful of inputs

## Stack trace
```
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
    at getRootForUpdatedFiber (/$bunfs/root/hunk:3318:131)
    at dispatchSetStateInternal (/$bunfs/root/hunk:4544:45)
    at dispatchSetState (/$bunfs/root/hunk:4515:31)
    at <anonymous> (/$bunfs/root/hunk:90041:17)
    at updateViewport (/$bunfs/root/hunk:90340:35)
    at handleViewportChange (/$bunfs/root/hunk:90346:21)
    at emit (node:events:95:22)
    at onChange (/$bunfs/root/hunk:67289:18)
    at value (/$bunfs/root/hunk:66967:23)
    at updateSliderFromScrollState (/$bunfs/root/hunk:67374:16)
```

A second occurrence (different mode, scroll-triggered) had a near-identical chain with one extra frame:
```
    at updateViewport (/$bunfs/root/hunk:90343:24)
    at handleViewportChange (/$bunfs/root/hunk:90346:21)
    at emit (node:events:95:22)
    at onChange (/$bunfs/root/hunk:67289:18)
    at value (/$bunfs/root/hunk:66967:23)
    at updateSliderFromScrollState (/$bunfs/root/hunk:67374:16)
    at scrollPosition (/$bunfs/root/hunk:67245:39)
```

## Likely cause
The chain is `slider state change → onChange → handleViewportChange → updateViewport (setState) → slider state change → ...`. Looks like a missing equality / source guard so a viewport update from the slider re-emits a slider update from the viewport on the same tick.

## Workaround
None layout-side — happens in `auto` and `stack` alike. Navigate slowly between hunks; restart Hunk after a crash. (`mode = "split"` not yet ruled out.)

### LINKED ISSUE #234: Bottom-edge scroll snap-back when agent notes are attached

### LINKED ISSUE #235: next/prev-comment keybinding jumps backward when selection is on a non-annotated hunk
## Summary
With agent notes attached, the next/prev-comment keybinding (and the equivalent `hunk session navigate --next-comment` / `--prev-comment` API) appears to "stop working" — pressing it teleports the user backward to the first comment instead of advancing to the next one. Repros in vanilla 0.10.0 once the user has scrolled (or otherwise moved selection) off any of the annotated hunks.

## Repro
1. `hunk diff <range>` on a multi-file diff
2. Attach agent notes to a few hunks
3. Press next-comment a couple of times to advance
4. Scroll a little (or press next-hunk) so the centered selection lands on a non-annotated hunk
5. Press next-comment again → viewport jumps back to the first comment instead of advancing to the next one

It's reliably reproducible because viewport-centered selection (`findViewportCenteredHunkTarget` → `onViewportCenteredHunkChange`) keeps shifting `selectedHunkIndex` to whichever hunk is centered as the user scrolls. That hunk is frequently *not* annotated, so the next press of next/prev-comment hits the no-exact-match fallback path.

## Root cause
`src/ui/lib/hunks.ts:findNextHunkCursor`:

```ts
const currentIndex = cursors.findIndex(
  (cursor) => cursor.fileId === currentFileId && cursor.hunkIndex === currentHunkIndex,
);
const nextIndex =
  currentIndex >= 0
    ? Math.min(Math.max(currentIndex + delta, 0), cursors.length - 1)
    : delta >= 0
      ? 0
      : cursors.length - 1;
```

When `(currentFileId, currentHunkIndex)` isn't an exact match in the (filtered) cursor list — which is the common case for annotated-cursor navigation, since the user's selection lives on the *full* hunk stream — the fallback unconditionally returns the first or last cursor regardless of where the user actually is. Often that's *backwards* from their position.

## Suggested fix
Resolve the relative position from the cursor list itself instead of falling back to first/last:

- Files appear in stream order in the cursor list, and within a file `hunkIndex` is ascending.
- The first appearance of each file in the cursor list gives a stable per-file stream rank.
- For a non-matching `(currentFileId, currentHunkIndex)`:
  - delta ≥ 0 → first cursor whose stream rank is strictly greater than the current position.
  - delta < 0 → last cursor whose stream rank is strictly less than the current position.
- Only fall back to start/end when the current file has no cursors at all (file with no annotations) or when the search exhausts the list.

The existing test cases at `src/ui/lib/hunks.test.ts:160-170` happen to coincide with the "first/last cursor" answer because they exercise a single-cursor list, so they don't catch the regression — worth adding regression tests for non-annotated current with a multi-cursor list.

## Related
- Different bug from #233 (max update depth) and #234 (bottom-edge snap-back with notes).
- Same broad area: viewport / selection bookkeeping under agent annotations.

kevinswiber:
## Summary

This is a combined fix for the agent-note navigation failures I hit while reviewing a live `--agent-context` session:

- keep agent-note geometry stable so total review height no longer changes as note-bearing files enter/leave the viewport
- resolve `{` / `}` and `hunk session navigate --next-comment` / `--prev-comment` relative to the full review stream when the current hunk is not itself annotated
- match annotation ranges against the full visible hunk span, including context rows, so saved `agent-context` anchors inside a hunk remain navigable
- keep comment navigation non-cyclic at annotated-stream edges, matching normal hunk-navigation clamp behavior

Fixes #234.
Fixes #235.
Refs #233.

## Relationship to the other open PRs

There are three related open PRs. This PR overlaps with them, but it is not identical.

- #242 is the direct fix for #233. It coalesces viewport listener updates with `queueMicrotask` to avoid setState during OpenTUI scrollbar/layout event emission. This PR does **not** include that direct viewport-loop guard, so it should not close #233. What this PR does for #233 is indirect: it removes the agent-note height oscillation that can feed the same update loop.
- #243 fixes #234 with a narrower `DiffPane` change: measure geometry with all notes, but keep rendering on the viewport-restricted note set. This PR fixes the same root cause by removing the split source entirely: geometry and rendered sections both receive `allAgentNotesByFile`. That is simpler and matches the one-source-of-truth direction, but it has a possible rendering-work tradeoff for visible/prefetched sections with many notes.
- #244 fixes one comment-navigation failure by widening `hunkLineRange` to use the full hunk header span. This PR includes that same category of fix, but also fixes a separate #235 path: when the current selected hunk is not annotated, `{` / `}` used to fall back to the first/last annotated hunk instead of resolving the next annotated hunk relative to the current full-stream position.

## What is worth keeping if the other PRs land first

If #242, #243, and #244 land before this PR, I can rebase and narrow this branch. The part that still appears worth merging from this PR is:

- stream-relative annotated navigation in `src/ui/lib/hunks.ts`
- the React/session callsites that pass the full hunk stream into that helper (`useReviewController.ts`, `reviewState.ts`)
- the unit regression in `src/ui/lib/hunks.test.ts`
- the live PTY regression for repeated `}` navigation through an `agent-context` sidecar

Those pieces are not reflected in #242, #243, or #244 as currently written.

Optional pieces after a rebase:

- The `DiffPane` geometry work should be dropped if maintainers prefer #243's narrower measurement-only approach.
- The `hunkLineRange` consolidation should be dropped if #244 lands first, unless maintainers prefer this branch's shared-helper shape.

### Bottom-edge snap-back with notes (#234)

1. Open a multi-file review: `hunk diff <range>`.
2. Attach agent notes to a file that can scroll off the top, for example:
   ```sh
   printf '%s\n' '{"comments":[{"filePath":"<early-file>","hunkNumber":1,"summary":"note"}]}' \
     | hunk session comment apply --repo <repo> --stdin
   ```
3. Scroll to the last line of the last file.
4. Try to scroll one line farther.

Before this change, the viewport can snap upward by roughly the height of off-top note rows because total content height changes with the viewport.

### Next/previous comment from a non-annotated hunk (#235)

1. Open a multi-file review and attach notes to several hunks.
2. Press `}` a couple of times to move through notes.
3. Scroll or press `]` so the selected/centered hunk is not one of the annotated hunks.
4. Press `}` again.

Before this change, the fallback path in `findNextHunkCursor` treats the current hunk as absent from the filtered annotated cursor list and jumps to the first annotated hunk instead of the next annotated hunk after the current stream position.

### Context-row anchors skipped by comment navigation

1. Use a hunk with leading context before the changed line.
2. Add a hunk-level comment so Hunk resolves the note anchor to the first changed row inside the visible hunk span.
3. Navigate with `}` or `hunk session navigate --next-comment`.

Before this change, `hunkLineRange` used `additionLines` / `deletionLines`, which count changed lines only, not the full hunk header extent. Anchors on context-adjacent rows could then fall outside the hunk range and never enter the annotated cursor list.

### Stable agent-note geometry

`DiffPane` now derives `sectionGeometry` from the full per-file note map rather than a viewport-restricted note map. That keeps `totalContentHeight` independent of `scrollViewport.top`, so bottom-edge clamping has a stable ceiling while scrolling.

### Stream-relative annotated navigation

`findNextHunkCursor` now accepts the full stream cursor list. If the current hunk is absent from the target subset, annotated navigation finds the nearest target cursor before/after the current stream position instead of falling back to the first/last annotated cursor.

When the current stream position is outside the annotated span, comment navigation clamps to the nearest annotated edge instead of wrapping. That matches normal hunk navigation and is now documented/tested explicitly.

Both the React review controller and session command path pass the full stream into that helper.

### Full hunk-span range matching

`hunkLineRange` now uses Pierre's required header span counts (`additionCount` / `deletionCount`). `agentAnnotations` uses that shared helper, so live comments, sidecar notes, and annotated-hunk discovery all agree on hunk extent.

### Review follow-up

After Greptile review, this branch drops the defensive keyboard callback refs rather than expanding that pattern. OpenTUI's current `useKeyboard` already keeps the latest handler via `useEffectEvent`, and the refs were not part of the root fix.

## Tradeoffs

- This intentionally removes the separate viewport-restricted note-measurement path. Rendered sections now receive the same full note set used for geometry. File-level windowing still limits which sections render, but a visible/prefetched section with many notes may do more per-render note placement work than #243's narrower measurement-only approach.
- This is a combined fix across related failure modes. If maintainers prefer the split upstream PRs, I can rebase this branch after they land and keep only the stream-relative annotated-navigation pieces listed above.
- This does not include the microtask coalescing approach from #242, so #233 should stay assigned to #242 or a similar direct viewport-loop fix.

## Testing

Passed on current branch head:

```sh
bun test src/core/liveComments.test.ts src/ui/lib/hunks.test.ts src/ui/components/ui-components.test.tsx src/ui/AppHost.interactions.test.tsx -t "line numbers|hunk ranges|overlapping annotations|uses full stream position|manual scrolling move away|comment"
tmp=$(mktemp -d); XDG_CONFIG_HOME="$tmp" bun test test/pty/ui-integration.test.ts -t "comment navigation between agent-context notes"; rc=$?; rm -rf "$tmp"; exit $rc
bun run lint
bun run format:check
```

`bun run typecheck` currently fails on `main` itself, independent of this branch:

```text
src/ui/AppHost.interactions.test.tsx(267,10): error TS2552: Cannot find name 'createTestGitAppBootstrap'. Did you mean 'createTestVcsAppBootstrap'?
```

I left that unrelated base issue out of this PR. I can rebase once `main` is fixed.

### DIFF
--- a/src/core/liveComments.test.ts
+++ b/src/core/liveComments.test.ts
@@ -38,7 +38,9 @@ describe("live comment helpers", () => {
     const file = createExampleDiffFile();
 
     expect(findHunkIndexForLine(file, "old", 1)).toBe(0);
+    expect(findHunkIndexForLine(file, "old", 3)).toBe(0);
     expect(findHunkIndexForLine(file, "new", 2)).toBe(0);
+    expect(findHunkIndexForLine(file, "new", 4)).toBe(0);
     expect(findHunkIndexForLine(file, "new", 40)).toBe(-1);
   });
 
@@ -111,8 +113,8 @@ describe("live comment helpers", () => {
     const range = hunkLineRange(file.metadata.hunks[0]!);
 
     expect(range.oldRange[0]).toBeLessThanOrEqual(1);
-    expect(range.oldRange[1]).toBeGreaterThanOrEqual(2);
+    expect(range.oldRange[1]).toBeGreaterThanOrEqual(3);
     expect(range.newRange[0]).toBeLessThanOrEqual(1);
-    expect(range.newRange[1]).toBeGreaterThanOrEqual(2);
+    expect(range.newRange[1]).toBeGreaterThanOrEqual(4);
   });
 });
--- a/src/core/liveComments.ts
+++ b/src/core/liveComments.ts
@@ -8,12 +8,16 @@ export interface ResolvedCommentTarget {
   line: number;
 }
 
-/** Compute the inclusive old/new line spans covered by one hunk. */
+/** Compute the inclusive old/new line spans covered by one hunk, including context rows. */
 export function hunkLineRange(hunk: Hunk) {
-  const additionSpan = hunk.additionCount ?? hunk.additionLines;
-  const deletionSpan = hunk.deletionCount ?? hunk.deletionLines;
-  const newEnd = Math.max(hunk.additionStart, hunk.additionStart + Math.max(additionSpan, 1) - 1);
-  const oldEnd = Math.max(hunk.deletionStart, hunk.deletionStart + Math.max(deletionSpan, 1) - 1);
+  const newEnd = Math.max(
+    hunk.additionStart,
+    hunk.additionStart + Math.max(hunk.additionCount, 1) - 1,
+  );
+  const oldEnd = Math.max(
+    hunk.deletionStart,
+    hunk.deletionStart + Math.max(hunk.deletionCount, 1) - 1,
+  );
 
   return {
     oldRange: [hunk.deletionStart, oldEnd] as [number, number],
--- a/src/core/liveComments.ts
+++ b/src/core/liveComments.ts
@@ -8,16 +8,12 @@ export interface ResolvedCommentTarget {
   line: number;
 }
 
-/** Compute the inclusive old/new line spans touched by one hunk. */
+/** Compute the inclusive old/new line spans covered by one hunk. */
 export function hunkLineRange(hunk: Hunk) {
-  const newEnd = Math.max(
-    hunk.additionStart,
-    hunk.additionStart + Math.max(hunk.additionLines, 1) - 1,
-  );
-  const oldEnd = Math.max(
-    hunk.deletionStart,
-    hunk.deletionStart + Math.max(hunk.deletionLines, 1) - 1,
-  );
+  const additionSpan = hunk.additionCount ?? hunk.additionLines;
+  const deletionSpan = hunk.deletionCount ?? hunk.deletionLines;
+  const newEnd = Math.max(hunk.additionStart, hunk.additionStart + Math.max(additionSpan, 1) - 1);
+  const oldEnd = Math.max(hunk.deletionStart, hunk.deletionStart + Math.max(deletionSpan, 1) - 1);
 
   return {
     oldRange: [hunk.deletionStart, oldEnd] as [number, number],
--- a/src/ui/components/panes/DiffPane.tsx
+++ b/src/ui/components/panes/DiffPane.tsx
@@ -40,7 +40,6 @@ import { VerticalScrollbar, type VerticalScrollbarHandle } from "../scrollbar/Ve
 import { prefetchHighlightedDiff } from "../../diff/useHighlightedDiff";
 
 const EMPTY_VISIBLE_AGENT_NOTES: VisibleAgentNote[] = [];
-const EMPTY_VISIBLE_AGENT_NOTES_BY_FILE = new Map<string, VisibleAgentNote[]>();
 
 /**
  * Clamp one vertical scroll target into the currently reachable review-stream extent.
@@ -270,8 +269,9 @@ export function DiffPane({
     return next;
   }, [files, showAgentNotes]);
 
-  // Keep exact row rendering for wrapped lines and the selected file's visible notes;
-  // other files can still use placeholders and viewport windowing.
+  // Keep exact row rendering for wrapped lines; other files can still use placeholders and viewport
+  // windowing. Rendered sections always include their note rows so their concrete height matches the
+  // geometry used for scroll planning.
   const windowingEnabled = !wrapLines;
   const [scrollViewport, setScrollViewport] = useState({ top: 0, height: 0 });
   const scrollbarRef = useRef<VerticalScrollbarHandle>(null);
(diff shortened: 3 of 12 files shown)


benvinegar:
Thanks for the detailed combined fix and write-up here. I’m going to close this as superseded rather than ask for another rebase, because the underlying pieces have since landed on main in smaller PRs:

- #242 fixed the rapid viewport/update-depth path.
- #243 stabilized agent-note geometry by measuring against the full note set.
- #244 switched hunk range matching to header counts so context-row anchors stay in range.
- #315 added stream-position-based comment navigation.
- #317 unified/expanded inline note behavior and coverage.

The issues this PR targeted (#233, #234, and #235) are now closed, and the remaining branch has conflicts against the newer note/navigation code. If there is still a missing edge case from this branch after fresh main, let’s bring it back as a narrow follow-up with a focused regression test.

PR #259, opened 2026-05-09
fix(git): pass --no-ext-diff when diffing untracked files

iamken1204:
## Summary

If you have `diff.external` set in git config (difftastic, delta, etc.), `hunk diff` crashes on any untracked file:

```bash
$ hunk diff
hunk: Expected one parsed file for untracked patch "foo.txt", got 0.
```

hunk forms untracked patches via `git diff --no-index`, which honors `diff.external` and outputs whatever that tool prints instead of unified diff. The regular diff/show paths already use `--no-ext-diff`, the untracked path was missing it.

## Test

- Added a regression that sets `diff.external` on a temp repo and loads an untracked file.
- Repro'd locally with `diff.external = difft`; works after the fix.

### DIFF
--- a/src/core/git.test.ts
+++ b/src/core/git.test.ts
@@ -1,7 +1,16 @@
 import { describe, expect, test } from "bun:test";
-import { runGitText } from "./git";
+import { buildGitStashShowArgs, runGitText } from "./git";
 
 describe("git command helpers", () => {
+  test("disables external diff tools for stash patches", () => {
+    const args = buildGitStashShowArgs({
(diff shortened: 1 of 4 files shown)


benvinegar:
Addressed the Greptile note by making `hunk stash show` pass `--no-ext-diff` too, so all Git-backed patch-producing paths now explicitly avoid external diff tools. I also adjusted the existing untracked-file regression to avoid the Unix-only `/usr/bin/true` command.

PR #262, opened 2026-05-09
feat(ui): Add y shortcut to yank selection to clipboard

amix:
## Context
Press `y` to copy the active terminal text selection to the clipboard. Uses `pbcopy` on macOS and falls back to OSC52 elsewhere. Action also surfaces in the Navigate menu and the help dialog.

## What was changed
- New `y` keybinding wired through `useAppKeyboardShortcuts`, the Navigate menu, and the help dialog
- New `src/ui/lib/clipboard.ts` with `yankActiveSelection` (selection read + copy + status mapping) and `copyTextToClipboard` (pbcopy → OSC52 fallback)
- New `useTransientNotice` hook hoisted to `AppHost` so yank acknowledgements and the startup update notice share one status-bar slot and one dismiss timer instead of competing
- `useStartupUpdateNotice` refactored to publish through the shared channel

## Out of scope
- No clipboard read; yank only copies what the user has selected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Added
 
+- Added a `y` yank shortcut for copying the current text selection to the clipboard.
+
 ### Changed
 
 ### Fixed
--- a/src/ui/App.tsx
+++ b/src/ui/App.tsx
@@ -23,6 +23,7 @@ import { useHunkSessionBridge } from "./hooks/useHunkSessionBridge";
 import { useMenuController } from "./hooks/useMenuController";
 import { useReviewController } from "./hooks/useReviewController";
 import { buildAppMenus } from "./lib/appMenus";
+import { yankActiveSelection } from "./lib/clipboard";
 import { fileRowId } from "./lib/ids";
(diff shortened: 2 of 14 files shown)


benvinegar:
Hey can you help me understand the use case? With iTerm or Ghostty I just drag-select and copy using the native terminal.

stefanprodan:
> Hey can you help me understand the use case? With iTerm or Ghostty I just drag-select and copy using the native terminal.

I find it very annoying that Cmd+C does not work. What's the point of allowing text selection if it's not possible to copy it?

benvinegar:
> I find it very annoying that Cmd+C does not work.

Okay sounds like we just need to fix that then. Maybe related to #312.

benvinegar:
Closing this as superseded by the copy-selection work that landed in #306. The original copying gap tracked in #312 is now closed: diff/pager views support mouse-drag selection and clipboard copy via OSC 52, with copy-decoration handling in the current render path.

A future keyboard yank command could still be proposed separately, but this branch predates the current copy-selection implementation and would need a fresh design on top of it.

benvinegar:
Reopening — #306 fixed mouse-drag copy, but it did not add the keyboard `y` yank shortcut this PR proposes. Keeping this open to evaluate the keyboard-driven copy workflow separately.

PR #267, opened 2026-05-09
fix(loaders): normalize mnemonic pager prefixes

### LINKED ISSUE #265: hunk pager fails with git diff when diff.mnemonicPrefix = true (no output or "no files match the filter")
## Issue

When hunk is configured as the pager for `git diff` (via `pager.diff = hunk pager` or similar), `git diff` often produces no output or the error "no files match the filter".

Running `hunk diff` directly works fine and shows the expected diff.

## Steps to reproduce

Add this to your `~/.gitconfig` (or equivalent)
```ini
[diff]
    mnemonicPrefix = true
```
Configure hunk as pager (example)
```ini
[core]
    pager = hunk pager
```
Run `git diff` (on a repo with changes).

## Expected behavior
`git diff` should pipe the diff to `hunk pager` and display it nicely (with sidebar, etc., as in recent releases).

## Actual behavior

- Often blank output
- Or hunk complains that no files match the filter
- `hunk diff` works as a workaround

## Additional context / Workaround

Commenting out / setting `diff.mnemonicPrefix = false` makes `hunk pager` work again.
This setting changes diff prefixes from the standard `a/` / `b/` to more informative ones like `i/` (index), `w/` (worktree), `c/` (commit), etc. Many tools that parse git diffs assume the default prefixes, so this is likely a parsing incompatibility.

## Environment

hunk version: 0.11.0
OS: macOS
Git version: 2.54.0
Terminal: Ghostty

## Screenshots

Image

benvinegar:
## Summary

Fixes `hunk pager` / `hunk patch` handling for Git patch input emitted with `diff.mnemonicPrefix=true`.

Git emits mnemonic side prefixes like `i/`, `w/`, `c/`, `1/`, and `2/` instead of the parser-friendly `a/` / `b/` prefixes. After #240, those patches can parse, but file paths keep the side prefix (for example `w/example.ts`). This normalizes mnemonic prefix pairs back to Pierre's expected `a/... b/...` shape before parsing so the review path is `example.ts`.

Fixes #265.

## Tests

- [x] `bun test src/core/loaders.test.ts`
- [x] `bun run typecheck`
- [x] `bun run lint`

*This PR description was generated by Pi using OpenAI GPT-5*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Fixed
 
+- Fixed `hunk pager` parsing for Git diffs emitted with `diff.mnemonicPrefix=true` so file paths do not keep `i/`, `w/`, `c/`, `1/`, or `2/` side prefixes.
 - Fixed large tracked and untracked file handling so very large diffs render as skipped placeholders instead of slowing startup or overflowing the JavaScript call stack.
 
 ## [0.11.0] - 2026-05-09
--- a/src/core/loaders.test.ts
+++ b/src/core/loaders.test.ts
@@ -1082,6 +1082,88 @@ describe("loadAppBootstrap", () => {
     expect(bootstrap.changeset.files[0]?.stats.additions).toBe(1);
   });
 
+  test("loads patch text emitted with diff.mnemonicPrefix=true (e.g. from `hunk pager` stdin)", async () => {
+    const dir = createTempRepo("hunk-patch-mnemonic-prefix-");
+
+    writeFileSync(join(dir, "example.ts"), "export const value = 1;\n");
+    git(dir, "add", ".");
+    git(dir, "commit", "-m", "initial");
+
+    writeFileSync(join(dir, "example.ts"), "export const value = 2;\n");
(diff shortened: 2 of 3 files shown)


benvinegar:
@greptile-apps thanks for the review — addressed all three points in `d2af79d`.

- The mnemonic-prefix detection now has access to the whole `diff --git` block and checks `rename from` / `rename to` metadata before stripping `c/`, `w/`, etc., so noprefix renames like `c/foo.ts -> w/bar.ts` keep those as real directory names.
- `splitGitMnemonicPrefix` now uses the prefix set directly instead of duplicating the character class plus a redundant `.has()` guard.
- Added regression tests for both mnemonic rename diffs and the noprefix mnemonic-looking-directory rename false positive.

benvinegar:
Thanks — small nuance on that `rewriteMode: "add"`: it is dead for the unquoted canonical path because `rewriteGitDiffHeader` returns `rewriteMode: null` from the `isCanonical` branch. It is still intentionally consumed through the quoted-header path, where `canonicalizeGitPathPair()` can receive already-prefixed quoted input like `"a/foo" "b/foo"`; we still need a non-null rewrite mode there so the matching quoted `---` / `+++` lines are canonicalized into Pierre’s unquoted form.

Given that, I’m going to leave it as-is rather than add a type-only refactor in this PR.

PR #277, opened 2026-05-10
feat(ui): mark files to hide them from the review stream

aldevv:
## Why

Long changesets mix real changes with noise (lockfiles, snapshots, vendored fixtures), and once you've reviewed a file there was no way to set it aside. You scroll past it every time and lose track of what you've already looked at.

## User-visible
image

`m` marks the focused file; `M` (shift+m) clears every active mark. Marked files disappear from the diff stream but stay in the sidebar dimmed and crossed out, with a `<n> hidden` footer hint reminding you how many are tucked away. Clicking a marked sidebar row unmarks it instead of trying to jump to a file that isn't there.

The legacy hunk-metadata toggle moved off `m` (now `H`) so the mark binding lives in the most reachable spot.

## What it does

- `useReviewController` owns the `markedFileIds` set and exposes `toggleMarkedFile`, `clearMarkedFiles`, `hiddenByMarkCount`, and an `unmarkedFiles` list alongside the existing `allFiles`/`visibleFiles`.
- Visibility derivation moves into `buildReviewState`, so the review stream, sidebar entries, and stats all read from one source. Subsequent filter and search features can layer on top of `unmarkedFiles` without re-deriving visibility ad hoc.
- `buildSidebarEntries` carries a `marked` flag per entry. `FileListItem` reads it to dim the icon/name and apply `STRIKETHROUGH`, so marked rows stay visible but obviously inert.
- `SidebarPane` renders a 1-row footer (`<n> hidden`) when any file is marked. Clicking a marked row routes through a new `onUnmarkFile` prop instead of `onSelectFile`.
- Help dialog and `appMenus` gain `Mark file (m)` / `Clear marks (Shift+M)` entries; the menu hook wires the keyboard shortcuts to the focused file.

## Tests

- `bun test src/ui/lib/reviewState.test.ts` — marked ids drop files from `unmarkedFiles`/`visibleFiles`, surface in `hiddenByMarkCount`, and the sidebar entries keep the marked file with `marked: true`.
- `bun test src/ui/lib/files.test.ts` — `buildSidebarEntries` flags marked rows and tolerates unknown ids.
- `bun test src/ui/hooks/useReviewController.test.tsx` — `toggleMarkedFile` round-trips, `clearMarkedFiles` is a no-op when empty, and the controller's `hiddenByMarkCount`/`unmarkedFiles` track the set.
- `bun test test/pty/marked-files-integration.test.ts` — new PTY test: `m` removes the focused file from the stream, the sidebar footer reports `1 hidden`, and `M` restores it.
- `bun test src/ui/AppHost.interactions.test.tsx`, `src/ui/components/ui-components.test.tsx`, `src/ui/lib/ui-lib.test.ts` — updated for the new `H` binding, sidebar footer rendering, and menu wiring.
- `bun run typecheck` and `bun run lint` clean.


benvinegar:
@aldevv I apologize being so slow for this. I want to acknowledge this is a good idea and requested by many people. I hesitated to respond to this because I wasn't sure I wanted to evolve `hunk` much past a diff tool.

I'm still unsure if this is the right approach, but I will mention it is not achievable using the extenson system (e.g. you can replace the existing pane or add a separate pane).

https://www.hunk.dev/docs/extend/extensions/

Anyways, I'm really sorry about the super slow response.

PR #278, opened 2026-05-10
fix(session): preserve live comments across daemon-driven reloads

### LINKED ISSUE #212: Reloading diff wipes agent review annotations
After performing a review, the agent annotates the diff with hunk-level comments. However, when I go through the diff to fix the issues, the diff reloads and the agent annotations from the current review disappear.

Expected behavior
Agent comments should persist while I’m reviewing and fixing the current diff.

Actual behavior
Reloading the diff clears/wipes the agent comments, making it hard to track and address the review feedback.

Impact
This breaks the review flow because I lose the annotations I’m actively working through.

mvanhorn:
## Summary

`hunk session reload` (the daemon-driven reload path) was wiping the in-memory live-comment state. After this PR, daemon reloads preserve live comments the same way the `r` keystroke and watch-mode auto-reload already do.

## Why this matters

Per #212, after an agent annotates the diff with hunk-level comments, calling `hunk session reload` (from a second terminal or from another agent message on the same session) reloaded the diff and dropped every annotation the user was actively reviewing. The maintainer confirmed the behavior in [the issue comment](https://github.com/modem-dev/hunk/issues/212#issuecomment-4330296427): "Thanks for sharing. This is definitely a problem."

The root cause is in `src/hunk-session/bridge.ts:70`: the `reload_session` handler forwarded `sourcePath` but not `resetApp`. `AppHost.tsx:63` treats undefined `resetApp` as truthy (`options?.resetApp !== false` is `true`), which increments `appVersion`, remounts `<App />`, and discards the React state in `useReviewController` -- including `liveCommentsByFileId`. The interactive `r` reload and watch-mode reload both pass `resetApp: false` (see `App.tsx:354`), so they preserve state; only the daemon path didn't.

## Changes

- `src/hunk-session/bridge.ts`: the `reload_session` handler now passes `resetApp: false` alongside `sourcePath`, matching the contract `App.tsx:354` already uses. The handler type now also accepts `resetApp` so future opt-in-to-hard-reset is a one-line addition.
- `src/hunk-session/bridge.test.ts`: extended the existing dispatch test with an exact-match assertion via `toHaveBeenCalledWith({ resetApp: false, sourcePath: "/repo" })` -- exact match so accidental regression to the old `undefined`-passing behavior is caught.
- `src/ui/AppHost.interactions.test.tsx`: new higher-level test that writes a live comment, dispatches a `reload_session` after rewriting the right-hand file, and asserts both the new content and the original review note appear in the rendered frame.

## Testing

- `bun run typecheck` -- clean
- `bun run format:check` -- clean
- `bun run lint` -- clean (0 warnings)
- `bun test ./src/hunk-session/bridge.test.ts ./src/ui/AppHost.interactions.test.tsx` -- 45 pass, 0 fail
- Manual smoke: start `hunk diff` in one terminal, write a live comment via the session bridge, run `hunk session reload` from a second terminal -- the comment persists across the reload.

Residual risk: an agent that explicitly wanted a hard reset on `reload_session` loses that capability. The handler type now accepts `resetApp`, so an explicit opt-in wire flag is a follow-up if anyone needs it.

Fixes #212

AI was used for assistance.

### DIFF
--- a/src/hunk-session/bridge.test.ts
+++ b/src/hunk-session/bridge.test.ts
@@ -116,7 +116,11 @@ describe("createHunkSessionBridge", () => {
       type: "command",
       requestId: "reload-1",
       command: "reload_session",
-      input: { sessionId: "session-1", nextInput: { kind: "vcs", staged: false, options: {} } },
+      input: {
+        sessionId: "session-1",
+        nextInput: { kind: "vcs", staged: false, options: {} },
+        sourcePath: "/repo",
+      },
     });
     await bridge.dispatchCommand({
       type: "command",
@@ -133,6 +137,10 @@ describe("createHunkSessionBridge", () => {
 
     expect(handlers.navigateToLocation).toHaveBeenCalledTimes(1);
     expect(handlers.reloadSession).toHaveBeenCalledTimes(1);
+    expect(handlers.reloadSession).toHaveBeenCalledWith(
+      { kind: "vcs", staged: false, options: {} },
+      { resetApp: false, sourcePath: "/repo" },
+    );
     expect(handlers.removeLiveComment).toHaveBeenCalledTimes(1);
     expect(handlers.clearLiveComments).toHaveBeenCalledTimes(1);
   });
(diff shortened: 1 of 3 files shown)


benvinegar:
Thanks for submitting. Seems reasonable. I just want to get the current version out there then I'll take a closer look.

mvanhorn:
No rush, thanks for the heads up @benvinegar. I'll keep the branch fresh -- ping me whenever you're ready.

PR #287, opened 2026-05-11
fix: clip CJK split rows by terminal cell width

iamken1204:
Fixes split-view rendering for long CJK lines, especially untracked files with Mandarin content, where mouse-wheel repainting could let wide characters spill past the right split pane into the left side.

Changes:
- Measure diff code and UI text by terminal cell width instead of UTF-16 string length.
- Slice, pad, wrap, and horizontally scroll styled diff spans without splitting fullwidth characters.
- Add Mandarin-only long-line regression coverage for mouse-wheel split-view repainting.
- Add Japanese and Korean unit coverage for CJK width and clipping behavior.

Before
CleanShot 2026-05-11 at 15 39 40@2x

After
CleanShot 2026-05-11 at 15 39 59@2x

### DIFF
--- a/src/ui/AppHost.scroll-regression.test.tsx
+++ b/src/ui/AppHost.scroll-regression.test.tsx
@@ -34,6 +34,27 @@ function createScrollBootstrap(): AppBootstrap {
   });
 }
 
+function createCjkUntrackedScrollBootstrap(): AppBootstrap {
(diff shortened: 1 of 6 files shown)


benvinegar:
🙏 Thank you for submitting, but rebasing cleanly was tough so I implemented in #397. Should be fixed in 0.15.x.

PR #289, opened 2026-05-11
feat(ui): add g / G navigation in hunk diff

### LINKED ISSUE #286: Add gg / G navigation in hunk diff
In Vim / git diff `gg` and `G` are used for quick top bottom jumps. Hunk already supports that via Home/End shortcuts. I get that we can learn another shortcut, but it would make switching from git diff to hunk diff much smoother for people like me, who already have it in their muscle memory. 

I'm proposing just 2 additions:
- `Home` and `gg` both jump to top
- `End` and `G` both jump to bottom

Happy to make a PR if it's approved

SalzDevs:
## Summary

- add less-style navigation aliases to Hunk review controls:
  - `g` jumps to top (alias for Home)
  - `G` jumps to bottom (alias for End)
- support the aliases in both full app mode and pager mode
- update controls help copy to include the new aliases
- add interaction coverage for `G` -> bottom and `g` -> top behavior
- update page-key interaction tests to send real PageUp/PageDown escape sequences so lowercase `g` does not interfere with test-only key-name strings

## Testing

- `bun run format:check`
- `bun run typecheck`
- `bun run lint`
- `bun test src/ui/AppHost.interactions.test.tsx src/ui/components/ui-components.test.tsx`

Closes #286

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Added
 
-- Added `gg` and `G` keyboard aliases for jump-to-top and jump-to-bottom review navigation.
+- Added `g` and `G` keyboard aliases for jump-to-top and jump-to-bottom review navigation.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Added
 
+- Added `gg` and `G` keyboard aliases for jump-to-top and jump-to-bottom review navigation.
+
 ### Changed
 
 ### Fixed
--- a/src/ui/AppHost.interactions.test.tsx
+++ b/src/ui/AppHost.interactions.test.tsx
@@ -18,6 +18,9 @@ import { createTestDiffFile as buildTestDiffFile, lines } from "../../test/helpe
 const { loadAppBootstrap } = await import("../core/loaders");
 const { AppHost } = await import("./AppHost");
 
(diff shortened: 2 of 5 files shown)


benvinegar:
What if we just mirrored `less` and did `g` / `G`? (vs `gg`)

SalzDevs:
Follow-up (previous comment lost code formatting due shell escaping):

- Rebased this branch on current `main` (resolved changelog conflict).
- Addressed the two Greptile notes:
  - added modifier guards to the `key.sequence === "G"` branch
  - added pager-mode interaction coverage for `G` (bottom) and `gg` (top)

For the `g` / `G` question: I kept `gg` + `G` to match #286 as written, but I can switch to strict less-style `g` + `G` if you prefer.

benvinegar:
> I can switch to strict less-style g + G if you prefer.

Yes please. Then we'll merge.

benvinegar:
Done — pushed a follow-up commit switching the shortcut to less-style `g` / `G`, updating help text, changelog, and the app/pager interaction coverage.

*This comment was generated by Pi using OpenAI GPT-5.1 Codex*

SalzDevs:
Updated after the less-style `g` / `G` follow-up:

- rebased the branch on current `main`
- preserved the maintainer follow-up commit switching to `g` / `G`
- added a small test-only fix so PageUp/PageDown interaction tests send real escape sequences instead of raw strings containing `g`
- refreshed the PR title/body to match the final behavior

Validation:

- `bun run format:check`
- `bun run typecheck`
- `bun run lint`
- `bun test src/ui/AppHost.interactions.test.tsx src/ui/components/ui-components.test.tsx`

benvinegar:
Thanks glad we got it sorted!

PR #303, opened 2026-05-13
feat(ui): add fold toggle for unchanged lines between hunks

adit-chandra:
## problem

I often want to inspect surrounding unchanged lines while reviewing diffs to better understand changes in context.
Common editors, review tools etc, including GitHub + Codex, already support expanding folded context inline.
This is the main missing feature preventing me from switching to Hunk as my primary code review tool.

## solution + design

- Expand folded context in place, not in a separate source view, so reviewers stay in the same file/hunk flow with agent notes, sticky headers, and scroll behavior intact.
- Treat full source text as a lazy per-file capability. Loaders attach exact old/new source endpoints only when they can identify the correct Git ref, index, worktree, file-pair, difftool, or untracked source.
- Avoid guessing. Raw patches and unsupported Git range shapes get no expander rather than risking context from the wrong revision.
- Convert expanded context into normal diff rows before render planning, so rendering, row windowing, gutter width, geometry, and mouse/keyboard interaction all use the same row model.
- Reuse Pierre/Shiki highlighting for loaded source lines, so expanded unchanged rows match the surrounding diff instead of rendering as plain text.
- Let the review controller own expansion state and source request lifecycles, which gives one place to invalidate soft reloads, ignore stale async completions, and show explicit error rows.
- Suggested review order: core source endpoints, then row synthesis/geometry, then controller/UI wiring.

## test plan

Added or updated coverage for:

- Source resolution: Git refs, index/worktree sides, file-pair diffs, difftool, show/stash, untracked files, deleted/renamed files, and staged additions before the first commit.
- Source failure handling: missing source, unexpected Git/read failures, caching, rejected fetchers, and visible error state.
- Row expansion: split/stack synthesized rows, trailing and leading gaps, deleted-file old-side expansion, CRLF/tab handling, syntax-aware spans, and keyboard gap selection.
- Layout/rendering: expanded rows affecting section height, later file placement, line-number gutter width, row geometry, clickable gap toggles, and syntax highlighting for expanded unchanged lines.
- Controller behavior: lazy load status, no-op without fetcher, reload invalidation, and stale pending source loads after soft reload.
- Shared fixtures: source fetcher and deferred-promise helpers to keep controller/component tests shorter and easier to scan.

Commands run:

- Passed: `bun run typecheck`
- Passed: `bun run format`
- Passed: `bun run format:check`
- Passed: `bun run lint`
- Passed: `git diff --check origin/main...HEAD`
- Passed: `git diff --check && git diff --cached --check`
- Passed: `bun test src/ui/diff/pierre.test.ts src/ui/diff/expandCollapsedRows.test.ts src/ui/components/ui-components.test.tsx`
- Passed: `bun test src/ui/hooks/useReviewController.test.tsx src/ui/components/ui-components.test.tsx src/core/git.test.ts src/core/fileSource.test.ts src/ui/diff/codeColumns.test.ts src/ui/diff/expandCollapsedRows.test.ts src/ui/lib/diffSectionGeometry.test.ts`
- Passed: `bun test src/core/loaders.test.ts -t "loads staged new files before the first commit|staged diffs against an explicit ref fetch old source from that ref|file-pair diffs attach a fetcher|git working-tree diffs read the new side|unstaged working-tree diffs read old source|hunk show <ref>|hunk stash show|untracked files attach a fetcher|deleted files attach a fetcher|renamed files read the old side|raw patch input does not attach"`
- Attempted: broader affected run including full `src/core/loaders.test.ts`; only the two upstream JJ loader tests failed because this machine does not have `jj` on `PATH`.

demo:

https://github.com/user-attachments/assets/c5f8871f-4fd1-4400-b3e2-66ea6a3f557b

### DIFF
--- a/src/ui/components/chrome/HelpDialog.tsx
+++ b/src/ui/components/chrome/HelpDialog.tsx
@@ -46,9 +46,8 @@ export function HelpDialog({
         ["1 / 2 / 0", "split / stack / auto"],
         ["s / t", "sidebar / theme"],
         ["a", "toggle AI notes"],
-        ["e", "toggle unchanged context"],
+        ["z", "toggle unchanged context"],
         ["l / w / m", "lines / wrap / metadata"],
-        ["e", "open file in $EDITOR"],
       ],
     },
     {
--- a/src/ui/components/chrome/HelpDialog.tsx
+++ b/src/ui/components/chrome/HelpDialog.tsx
@@ -46,6 +46,7 @@ export function HelpDialog({
         ["1 / 2 / 0", "split / stack / auto"],
         ["s / t", "sidebar / theme"],
         ["a", "toggle AI notes"],
+        ["e", "toggle unchanged context"],
         ["l / w / m", "lines / wrap / metadata"],
         ["e", "open file in $EDITOR"],
       ],
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable user-visible changes to Hunk are documented in this file.
 ### Added
 
 - Added mouse-drag text selection in diff views that copies selected rows to the system clipboard via OSC 52. A `View > Copy decorations` toggle (or `copy_decorations` config) controls whether the clipboard includes diff rails, gutters, and file headers or only the changed code.
+- Added inline expansion for collapsed unchanged file content. Click an unchanged-context row (`▾ N unchanged lines` when expandable, otherwise the static `··· N unchanged lines ···` form) or press `e` while a hunk is selected to reveal surrounding and trailing file lines without leaving the review. The affordance is shown only for input modes that have reachable source content (`hunk diff`, `show`, `stash show`, file-pair `diff` and `difftool`, untracked files); raw `hunk patch` input still renders as before. Failed and in-flight loads surface a one-line status ("Loading…", "Could not load N unchanged lines") on the gap row. Expanded context rows use the same syntax highlighting as the surrounding diff.
 
 ### Changed
 
--- a/src/core/diffFile.ts
+++ b/src/core/diffFile.ts
@@ -2,6 +2,7 @@ import { getFiletypeFromFileName, type FileDiffMetadata } from "@pierre/diffs";
 import { findAgentFileContext } from "./agent";
 import { patchLooksBinary } from "./binary";
 import { normalizeDiffMetadataPaths, normalizeDiffPath } from "./diffPaths";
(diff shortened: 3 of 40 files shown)


benvinegar:
@adit-chandra This is super awesome, I experimented w/ this briefly but didn't go all the way. Absolutely want this in the app.

My biggest concern is:
1) This is a huge PR
2) We are hardcoding a lot of git stuff as support for other VCSs is emerging (e.g. jujutsu and maybe Sapling, see #291)

So I want to hold off on this for a moment. I'm going to refactor the current code to abstract more over VCSs, then come back.

benvinegar:
@adit-chandra If you could rebase I can take another look. Those refactors are in place.

adit-chandra:
> @adit-chandra This is super awesome, I experimented w/ this briefly but didn't go all the way. Absolutely want this in the app.
> 
> My biggest concern is:
> 
> 1. This is a huge PR
> 2. We are hardcoding a lot of git stuff as support for other VCSs is emerging (e.g. jujutsu and maybe Sapling, see [feat: add Sapling (sl) VCS backend support #291](https://github.com/modem-dev/hunk/pull/291))
> 
> So I want to hold off on this for a moment. I'm going to refactor the current code to abstract more over VCSs, then come back.

hey @benvinegar, thanks for giving this a look! sorry for the late response on my end. 

Definitely agree, that this PR was larger than I'd like. I tried to keep it minimal, but given the cross-cutting nature, I wasn't able to reduce it's scope further.

In the future, would you prefer smaller, stacked PRs for review-ability? Or do you have other recs for improving review-ability of cross-cutting changes like this?

Happy to see you were able to move forward with the expansion toggles though :)

src/ui/components/chrome/HelpDialog.tsx
```diff
@@ -46,8 +46,8 @@ export function HelpDialog({
         ["1 / 2 / 0", "split / stack / auto"],
         ["s / t", "sidebar / theme"],
         ["a", "toggle AI notes"],
+        ["z", "toggle unchanged context"],
         ["l / w / m", "lines / wrap / metadata"],
-        ["e", "open file in $EDITOR"],
```
odjhey, line 50: was this intended to be removed, noticed it just now after upgrading to 0.15.0, can't find docs anywhere discussing the removal. thanks

odjhey: #402

duarteocarmo: ahaha - just noticed this as well.

PR #305, opened 2026-05-13
feat(ui): add Catppuccin themes

### LINKED ISSUE #304: Add Catppuccin themes
Add built-in [Catppuccin](https://github.com/catppuccin) Latte and Mocha themes so users can match Hunk with their terminal and editor color setup without custom theme overrides.

Implementation: #305

mfkd:
## Summary

- Add built-in Catppuccin Latte and Mocha themes.
- Wire the new theme ids through app and OpenTUI theme resolution.
- Update docs, changelog, and theme coverage tests.

Closes #304.

## Testing

- `bun run typecheck`
- `bun test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Added
 
+- Added Catppuccin Latte and Mocha as built-in themes.
 - Added mouse-drag text selection in diff views that copies selected rows to the system clipboard via OSC 52. A `View > Copy decorations` toggle (or `copy_decorations` config) controls whether the clipboard includes diff rails, gutters, and file headers or only the changed code.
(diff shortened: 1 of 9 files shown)


zeroaddresss:
I've set the theme in my dotfiles/hunk/config.toml file but it won't apply succesfully:
`theme = "catppuccin-mocha"   # graphite, midnight, paper, ember, catppuccin-latte, catppuccin-mocha, custom`

moreover, it does not even show as an option in the TUI:
image

mfkd:
> I've set the theme in my dotfiles/hunk/config.toml file but it won't apply succesfully: `theme = "catppuccin-mocha" # graphite, midnight, paper, ember, catppuccin-latte, catppuccin-mocha, custom`
> 
> moreover, it does not even show as an option in the TUI: image

@zeroaddresss  Your config looks right, but this feature isn't in a release yet — you'll need to test it locally against `main`:

```
bun install
bun run src/main.tsx -- diff --theme catppuccin-mocha
```

PR #306, opened 2026-05-13
feat(ui): add mouse-drag text selection and clipboard copy

raulk:
Hunk is very cool, but it's missing some basic features. Chipping in this PR to add selections via OpenTUI's OSC 52 support. 

Concretely, this adds:

**Mouse-drag text selection.** Click anywhere on a diff row and drag to select a range of text. The selection spans multiple rows when you drag across them, and a character-level visual highlight follows the pointer. Releasing the drag copies the selected text to your clipboard via OSC 52.

**Double-click and triple-click.** Double-click expands the selection to word boundaries within the clicked line. Triple-click copies the entire line.

**Split mode awareness.** Triple-click and drag selection automatically stay within the side you started on. When copy decorations are enabled, column offsets are computed relative to the correct pane, so what you see highlighted is exactly what lands on the clipboard.

**Copy decorations toggle.** A View > Copy decorations menu item (_off by default_, or set copy_decorations in config) controls whether copied text includes diff rails, line numbers, gutters, and file headers. When off, only the changed code text is copied.


benvinegar:
@raulk Very cool, played around w/ this, just want to test more.

raulk:
Yeah would appreciate more adversarial testing! I'm sure there's a bunch of edge cases lurking, because terminals are terminals. (I tested on Ghostty only)

PR #309, opened 2026-05-14
feat(embedded, ui): support embedding Hunk in host TUIs

khoaHyh:
## Summary

Playing with Hunk internals for an OpenCode plugin, and turned into a real embedded API instead of a one-off custom route.

- Add a public `hunkdiff/embedded` export for creating an embedded Hunk session and mounting it inside a host OpenTUI app.
- Support worktree, staged, VCS range, `show`, `stash show`, file diff, patch, and difftool sources through the same config + loader pipeline as the CLI.
- Scope renderer dimensions, resize events, and keyboard/paste input to the host-provided container so an embedded Hunk view can stay mounted without stealing input while inactive.
- Keep embedded sessions connected to the normal Hunk session broker, including launching the broker from the packaged `hunkdiff` binary.
- Centralize shared review command state so CLI sessions, broker commands, the UI controller, and embedded sessions use the same selected hunk, comments, notes, and snapshot behavior.

## Why

The original goal was to embed `hunkdiff` inside OpenCode's TUI. Rather than reaching into renderer internals from the plugin, this gives host TUIs a small supported surface:

- `createEmbeddedHunkSession(...)` owns source loading, reloads, snapshots, live comments, notes, and broker registration.
- `mountEmbeddedHunkApp(...)` renders the existing Hunk review UI into a host OpenTUI container.

## LLM use disclosure

I used gpt 5.5 & kimi k2.6 to develop this PR via opencode & pi.
I used gpt 5.5 to draft this PR description and then edited and filtered out some stuff myself.
All diffs were dogfooded with hunk inside opencode via the agent notes feature. Also used plannotator-review for the code tours feature for convenient comprehension.


benvinegar:
Am I misinterpreting, but does this mean maintaining an internal solidjs fork going forward?

Edit (after more than a passing glance): Okay, I guess it's not that bad. Just wondering how we can avoid drift between the React and Solid JS exports.

khoaHyh:
@benvinegar I didn't vibe with the internal SolidJS fork either 🤮 and I've since played around with a different approach

it now works through `src/embedded`:
- loads any supported Hunk source using same config and diff pipeline as the CLI
- mounts Hunk’s existing React/OpenTUI app into a renderer/container
- my plugin stays responsible for OpenCode commands, overlay lifecycle, focus, and key input scoping
- the embedded session registers with Hunk’s daemon/session broker, so agent notes, navigation, reloads, and comments still work even when the UI is not currently visible

just a bit of clean-up to do here then it should be ready!


https://github.com/user-attachments/assets/7126f4d5-0a90-4bd7-8d9c-22859a592e2b

khoaHyh:
Going to close this PR for now for two reasons:
- the performance of embedding Hunk in-process into OpenCode vs the new OpenCode native diff viewer is night and day. The native diff viewer is built directly into its own Solid/OpenTUI tree while the route I took made Hunk’s React tree and OpenCode’s host renderer both participate enough that the cost compounds and you experience this at large diffs
- I think a better direction would be to expose Hunk's agent notes capabilities in the new OpenCode diff viewer but I'll leave that for another PR

Was fun to play around here though, lots of great stuff in `hunk`. 

Attached is a demo of using Hunk's agent notes feature with the new native OpenCode diff viewer (might PR this in the future).

https://github.com/user-attachments/assets/641974f9-390a-4322-8e4a-ca480f3be603


https://github.com/user-attachments/assets/f406a23b-7a60-48bd-97f8-0be3e16bd8a3

PR #310, opened 2026-05-14
feat(ui): open file in editor with e shortcut

duarteocarmo:
<blockquote class="twitter-tweet"><p lang="en" dir="ltr">I want to avoid turning hunk into some whole git workflow thing<br><br>Open in editor seems reasonable if we’re talking $ EDITOR</p>&mdash; Ben Vinegar (@bentlegen) <a href="https://twitter.com/bentlegen/status/2054559913129497028?ref_src=twsrc%5Etfw">May 13, 2026</a></blockquote> 

## Summary
- add an `e` shortcut to open the currently selected sidebar file in `$EDITOR`
- open at the selected hunk line when available
- suspend the TUI while editing, then resume on editor exit
- auto-refresh the current diff/session after returning from the editor
- add menu/help entries: "Open file in editor"

I also ran the tests - and they passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Added
 
+- Added an `e` shortcut to open the selected diff file in `$EDITOR`.
 - Added `g` and `G` keyboard aliases for jump-to-top and jump-to-bottom review navigation.
 
 ### Changed
--- a/src/ui/App.tsx
+++ b/src/ui/App.tsx
@@ -24,6 +24,7 @@ import { useMenuController } from "./hooks/useMenuController";
 import { useReviewController } from "./hooks/useReviewController";
 import { buildAppMenus } from "./lib/appMenus";
 import { fileRowId } from "./lib/ids";
+import { openSelectedFileInEditor } from "./lib/openInEditor";
 import { resolveResponsiveLayout } from "./lib/responsive";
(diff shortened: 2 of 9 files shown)


duarteocarmo:
@benvinegar  - should be ready - lmk if anything you don't like.

benvinegar:
@duarteocarmo This is pretty handy. I just played with it.

But I notice there's a flash of content when you hit `e` that looks like some git diff output. Is it possible to avoid that?

duarteocarmo:
@benvinegar  - for what editor? terminal based or?

duarteocarmo:
@benvinegar  - ok. I made a slight change. 

- For non terminal-based editors, we don't need to suspend - so they just open and there's no flicker or anything. 
- For terminal based editors (vi, etc.) we actually need to call renderer suspend(), OpenTUI doesn't have an option to keep showing the alternate screen while doing so. (Other than some hacks like writing escape sequences to the termina - which I would honestly avoid)

lmk what you think

benvinegar:
Could you add tests for src/ui/lib/openInEditor.ts?

Also a Windows concern: it builds a shell command with Unix-style quoting + shell: true, which might not work w/ cmd.exe.

And please add a changelog entry.

duarteocarmo:
@benvinegar  - using `Bun.spawnSync` like the rest of the repo - which should not cause problems on most Windows setups. Added tests and Changelog entry. lmk if anything else missing!

PR #318, opened 2026-05-15
Preserve auto theme across watch refreshes

dive:
## Summary

- Preserve the startup-detected terminal theme mode across soft `--watch` refreshes
- Keep `theme: auto` from falling back to the default dark theme after watched file diffs reload

## Solves the bug

When Hunk started with `--theme auto`, startup detected the terminal background once & stored the resolved terminal theme mode on the initial app bootstrap.

In `--watch` mode, file changes trigger a soft reload: the diff bootstrap is replaced, but the `App` component is intentionally not remounted so view state like selected theme, layout, and scroll position can survive. That meant `themeId` stayed as `"auto"`, but the replacement bootstrap no longer had the startup-only detected terminal theme mode.

As a result, `resolveTheme("auto", null)` fell back to the default dark theme, so users on light terminals could see Hunk switch from the light `Paper` theme to dark `Graphite` after the first watched refresh.

This fix snapshots the startup-detected theme mode for the lifetime of the mounted app, so soft reloads keep resolving `auto` consistently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Fixed
 
+- Preserved the resolved auto theme across `--watch` refreshes instead of falling back to the default dark theme.
 - Included the bundled Hunk review skill in standalone prebuilt release archives so `hunk skill path` works after extracting a tarball or installing via Homebrew.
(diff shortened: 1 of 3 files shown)


benvinegar:
@dive Thanks!

PR #321, opened 2026-05-16
fix: statically link prebuilt Linux binaries against musl

### LINKED ISSUE #300: Pre-built Linux binaries should statically link against musl
Earlier PR https://github.com/modem-dev/hunk/pull/12 mentions this as a future thing to explore.

Right now the Linux binary builds are dynamically linked against whatever local linux environment the build happens in, which works on most distros but breaks for non-FHS distros (like NixOS, guix, etc).

For example:

```console
$ npx hunkdiff
Need to install the following packages:
hunkdiff@0.11.1
Ok to proceed? (y) y
Could not start dynamically linked executable: /home/agent/.npm/_npx/4c45ab9f0afbb831/node_modules/hunkdiff-linux-x64/bin/hunk
NixOS cannot run dynamically linked executables intended for generic
linux environments out of the box. For more information, see:
https://nix.dev/permalink/stub-ld
```

musl is a tiny standalone stdlib build target that can be statically linked against and will work on any linux distro, popular npm binaries like codex and opencode do this as well.

Not sure what is the best approach to integrate into the current build workflow. Perhaps something along the lines of `bun run build:release -DZIG_TARGET=x86_64-linux-musl  -DCMAKE_EXE_LINKER_FLAGS="-static"` (untested).

(Disclaimer: I am not an npm packaging or bun expert, this content is meat-written after some research so it may be incorrect.)

mvanhorn:
## Summary

Prebuilt Linux binaries from CI link against glibc, so users on Alpine/musl systems can't run them. Switch to a musl static-link target.

## Why this matters

This is the packaging fix issue #300 explicitly asks for. The blocker is purely the CI release workflow's link target; the source code is fully portable.

## Changes

- `.github/workflows/release.yml` - add a musl static-link target alongside the existing glibc target for Linux releases.
- README install snippet updated to mention musl compatibility.

Fixes #300

### DIFF
--- a/.github/workflows/release-prebuilt-npm.yml
+++ b/.github/workflows/release-prebuilt-npm.yml
@@ -98,8 +98,10 @@ jobs:
         include:
           - package_name: hunkdiff-linux-arm64
             runner: ubuntu-24.04-arm
+            target_triple: bun-linux-arm64-musl
           - package_name: hunkdiff-linux-x64
             runner: ubuntu-latest
+            target_triple: bun-linux-x64-musl
           - package_name: hunkdiff-windows-x64
             runner: windows-latest
           - package_name: hunkdiff-darwin-x64
@@ -119,6 +121,8 @@ jobs:
         run: bun install --frozen-lockfile
 
       - name: Build host artifact
+        env:
+          HUNK_TARGET_TRIPLE: ${{ matrix.target_triple }}
         run: |
           bun run build:bin
           bun run ./scripts/build-prebuilt-artifact.ts --expect-package "${{ matrix.package_name }}"
--- a/.github/workflows/release-prebuilt-npm.yml
+++ b/.github/workflows/release-prebuilt-npm.yml
@@ -34,8 +34,10 @@ jobs:
         include:
           - package_name: hunkdiff-linux-arm64
             runner: ubuntu-24.04-arm
+            target_triple: bun-linux-arm64-musl
           - package_name: hunkdiff-linux-x64
             runner: ubuntu-latest
+            target_triple: bun-linux-x64-musl
           - package_name: hunkdiff-windows-x64
             runner: windows-latest
           - package_name: hunkdiff-darwin-x64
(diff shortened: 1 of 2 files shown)


mvanhorn:
Rebased onto main (5b3d6bf). The conflict came from the build-bin.sh -> build-bin.ts rewrite in #285 and the new Windows matrix entry from the same PR. The musl-target conditional now lives in `scripts/build-bin.ts` (gated on `process.env.HUNK_TARGET_TRIPLE`) and the `target_triple` matrix entries are restricted to the two Linux runners.

On the Greptile P2 ("HUNK_TARGET_TRIPLE always injected for every matrix entry"): the `if (targetTriple)` guard in build-bin.ts already skips `--target` when the matrix entry doesn't set one, so macOS and Windows entries receive an empty string and produce a native build, which is the intended behavior. The two suggested mitigations would either skip the entire `Build host artifact` step on non-Linux (breaking those builds) or require splitting it into two near-duplicate steps to avoid passing through an empty env var that's already guarded one level down. Happy to revisit if a future consumer of `HUNK_TARGET_TRIPLE` lands without the same guard.

wuxiaobai24:
Hey @mvanhorn — still blocked by this on an older glibc host (`GLIBC_2.xx not found` from the npm prebuilt), so same root cause as #300 beyond just NixOS/Alpine.

Looks like this PR needs a rebase onto current `main` (merge conflicts) and a quick `oxfmt` pass on `scripts/build-bin.ts` (CI was red on format last run).

Would you be up for refreshing the branch? Happy to help test the musl artifact (`file`/`ldd` + run on an older glibc box) once CI is green. If you don’t have bandwidth, I can open a small follow-up PR that reuses your approach and credits #321.

mvanhorn:
Refreshed. Rebased onto current main (the conflict was the --no-compile-autoload-bunfig flag that landed in the meantime; the build now passes it alongside the musl --target logic) and ran oxfmt over scripts/build-bin.ts, so the format check should be green. Pushed as 19ef097.

Testing the musl artifact with file/ldd on an older glibc box once CI finishes would be a big help, thanks for the offer.

PR #325, opened 2026-05-17
fix(release): ignore staged npm artifact when archiving

benvinegar:
## Summary
- limit the GitHub release job download to `hunkdiff-*` platform artifacts
- filter the archive loop to platform artifact directories so staged npm release artifacts are ignored

## Why
The 0.13.0-beta.0 release workflow downloaded `staged-prebuilt-npm-release` alongside binary artifacts and tried to archive it as if it contained a top-level `hunk` binary.

## Tests
- `git diff --check`
- local shell simulation with an extra `staged-prebuilt-npm-release` directory

### DIFF
--- a/.github/workflows/release-prebuilt-npm.yml
+++ b/.github/workflows/release-prebuilt-npm.yml
@@ -184,6 +184,7 @@ jobs:
       - name: Download platform artifacts
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
+          pattern: hunkdiff-*
           path: dist/release/artifacts
 
       - name: Archive release assets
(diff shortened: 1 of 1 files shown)


benvinegar:
Note: the Windows compatibility check failed before any project commands ran, during `oven-sh/setup-bun` immediately after a Bun cache hit. It appears unrelated to this workflow-only release artifact filter change. I’ll rerun it once the workflow finishes.

benvinegar:
CI failures look unrelated/flaky:

- Windows compatibility failed in `oven-sh/setup-bun` before any project commands ran.
- Linux `Typecheck + Test + Smoke` failed in one PTY integration assertion (`explicit split mode stays split after a live resize`) where the terminal capture was truncated; this is unrelated to the release workflow YAML change.

I reran the failed jobs.

PR #334, opened 2026-05-19
test(ui): fix broken plannedReviewRows test on main

benvinegar:
## Summary

`main` currently fails `tsc` (and therefore the Typecheck + Test + Smoke CI job) independently of any open PR.

PR #317 (`feat: unify inline notes and expand behavior coverage`) unified inline notes into the diff-row stream:
- `inline-note` rows now require a `note: VisibleAgentNote` field
- the standalone `note-guide-cap` row kind was removed (a guide cap is now a `diff-row` carrying `noteGuideSide`)

But its colocated `src/ui/diff/plannedReviewRows.test.ts` still constructed the old shapes, so `tsc` fails on `main`:

```
plannedReviewRows.test.ts(81,3): error TS2322: Property 'note' is missing ... PlannedReviewRow
plannedReviewRows.test.ts(102,5): error TS2322: Type '"note-guide-cap"' is not assignable to type '"diff-row" | "inline-note"'
```

## Changes

Test-only. Update the two stale helpers to the current `PlannedReviewRow` model:
- `inlineNote()` now includes the required `note` field
- `guideCap()` now returns a `diff-row` with `noteGuideSide: "new"` (the unified representation)

No assertions changed — a guide cap is still a 1-height row that contributes to hunk bounds, so all three geometry tests pass with their original expected values.

## Testing

- `bun run typecheck` — clean
- `bun run lint` — clean
- `bun test src/ui/diff/plannedReviewRows.test.ts` — 3 pass

No CHANGELOG entry: test-only, no user-visible behavior change.

### DIFF
--- a/src/ui/diff/plannedReviewRows.test.ts
+++ b/src/ui/diff/plannedReviewRows.test.ts
@@ -91,20 +91,49 @@ function inlineNote(key: string, hunkIndex: number): PlannedReviewRow {
       summary: "Explain why this branch changed.",
       rationale: "The note should reserve space in the hunk bounds.",
     },
+    note: {
+      id: "note-1",
+      annotation: {
+        id: "note-1",
+        newRange: [1, 1],
+        summary: "Explain why this branch changed.",
+        rationale: "The note should reserve space in the hunk bounds.",
+      },
+    },
     anchorSide: "new",
     noteCount: 1,
     noteIndex: 0,
   };
 }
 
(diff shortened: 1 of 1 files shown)


benvinegar:
Superseded — #332 already fixed `plannedReviewRows.test.ts` on main as part of its changes, so this is no longer needed. Verified `bun run typecheck` is clean on current main.

PR #341, opened 2026-05-20
feat: add Kitty follow integration

reVrost:


reVrost:
@codex review

benvinegar:
Hey so this is a 1,000 line diff. Can you help me understand the goals? How you're using this?

This adds a Python script for example, which is a huge added dependency.

benvinegar:
Closing this. We don't want to embed and distribute a Python watcher as part of Hunk, and the motivating reason for making Kitty follow a core feature was never made clear enough to justify the added dependency and protocol surface. This would be better explored as an external extension if there is still interest.

PR #346, opened 2026-05-22
feat(diff): add a keyboard- and mouse-driven comment cursor

kertonandrew:
## Summary

Adds a comment cursor to the review stream so a human reviewer can drop notes inline next to the code they want an agent to look at, without leaving the TUI or learning a new CLI.

- **Keyboard**: `c` toggles cursor mode, `j`/`k` and `[`/`]` walk diff rows / hunks, `i` or Enter opens an inline composer, Enter saves, Esc cancels.
- **Mouse**: clicking any addition or deletion row positions the cursor and enters cursor mode, ready to compose. Clicks on context, headers, and empty cells are intentionally no-ops to match the keyboard navigation policy.
- **Persistence**: user notes flow through the same `liveCommentsByFileId` store as MCP comments. Agents pick them up unchanged via the existing `hunk session comment *` commands. No new protocol.
- **Presentation**: user-authored notes render with an orange "My note" palette across all four built-in themes so they read distinctly from purple "AI note" cards. The cursor row gets a row-wide tinted background; single-row inline notes drop the now-redundant trailing cap.
- **Viewport**: the cursor row auto-scrolls into view on every position change but not on pure mode toggles, so a tap of `c` doesn't yank the viewport.

The architectural rules from `CLAUDE.md` are preserved:

- Single source of truth: cursor state lives in `useReviewController` next to the existing selection state; the composer rides through the same `buildReviewRenderPlan` pipeline as inline notes; user comments use the same store as MCP comments.
- No parallel codepaths: when `commentCursor.mode === "off"` everything behaves exactly as before.
- Pane components stay focused: `App.tsx` orchestrates, `DiffPane` plumbs, `PierreDiffView` renders, helpers do the heavy logic.

CHANGELOG is updated under `## [Unreleased]`.

## Test plan

- [x] `bun run format:check`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test` — 537 pass / 5 pre-existing `jj`-not-in-PATH fails (CI installs jj)
- [x] `bun run test:integration` — 33 pass (PTY end-to-end coverage including a real `c` → `i` → type → Enter round trip)
- [x] `bun run test:tty-smoke` — 9 pass
- [x] Manual smoke through extended dogfooding: toggle, navigate, click, compose, autoscroll, multi-file scope all verified on a live diff
- [ ] CI: format check + lint + typecheck + tests + PTY + TTY smoke + npm pack/prebuilt checks + Windows compatibility job

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,18 @@ All notable user-visible changes to Hunk are documented in this file.
 ### Added
 
 - Added a keyboard-driven comment cursor (`c`) so reviewers can mark a diff row, type a single-line note with `i` / Enter, and persist it as a user-authored review comment that agents pick up through the existing `hunk session comment *` commands.
+- Added mouse support to the comment cursor: clicking any addition or deletion row positions the cursor and enters cursor mode, ready for `i` / Enter to open the composer.
+- Added an orange "My note" palette for user-authored review notes so they read distinctly from purple AI notes across all four built-in themes.
 
 ### Changed
 
+- Highlighted the entire diff row under the comment cursor (not just the left rail) and dropped the redundant trailing cap on single-row inline notes so the review stream reads cleaner.
+- Auto-scroll the review stream so the comment cursor stays in view when keyboard navigation crosses past the viewport edge.
+
 ### Fixed
 
+- Fixed the comment cursor painting in every file that happened to share the same hunk/side/line as the active cursor; it now appears only in the file the cursor is actually in.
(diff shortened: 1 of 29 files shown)


kertonandrew:
Closing — upstream v0.13.0 (#281, #317) already shipped session-persistent user-authored inline notes with the same `c` shortcut and CLI surface this PR was building toward. This branch's design (widened `LiveComment.source`) diverges from the upstream `SessionReviewNoteSummary` model, so the diff isn't a clean rebase target. Closing as superseded.

PR #347, opened 2026-05-22
fix: resolve editor paths from repo root

duarteocarmo:
@benvinegar - there is a bug when launching the editor from a place that is not the repo root. 
This PR should fix it.


benvinegar:
Makes sense, thanks for the follow up!

PR #355, opened 2026-05-23
Security: Unsafe shell execution with user-controlled pager command

tomaioo:
## Summary

Security: Unsafe shell execution with user-controlled pager command

## Problem

**Severity**: `High` | **File**: `src/core/pager.ts:L45`

The `pagePlainText` function in `src/core/pager.ts` spawns a pager command with `shell: true` using the `resolveTextPagerCommand` result. While the function attempts to prevent recursive `hunk` launches via regex, the `PAGER` or `HUNK_TEXT_PAGER` environment variable is passed directly to the shell without proper sanitization. An attacker who can control these environment variables could inject arbitrary shell commands (e.g., `PAGER='$(malicious)'` or backtick injection). The regex check `/(^|\s)hunk(\s|$)/` is insufficient to prevent command injection.

## Solution

Avoid `shell: true` and instead parse the pager command into arguments using a safe argument parser, or use `execFile` with an explicit array of arguments. If shell is required, strictly validate the pager command against an allowlist of known safe pagers, or use `shell: false` with `spawn` and pass the command as an first argument with any flags as subsequent array elements.

## Changes

- `src/core/pager.ts` (modified)

### DIFF
--- a/src/core/pager.ts
+++ b/src/core/pager.ts
@@ -24,10 +24,47 @@ export function resolveTextPagerCommand(env: NodeJS.ProcessEnv = process.env) {
   return candidate;
 }
 
+function parsePagerCommand(command: string): [string, string[]] {
+  const args: string[] = [];
+  let current = "";
+  let inQuotes = false;
+  let quoteChar = "";
+
+  for (let i = 0; i < command.length; i++) {
+    const char = command[i];
+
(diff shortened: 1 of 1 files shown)


benvinegar:
More exhaustive solution with tests in #358

PR #394, opened 2026-06-03
feat(ui): add Zenburn built-in theme

ramin:
## Summary

Adds the classic **Zenburn** (https://en.wikipedia.org/wiki/Wikipedia:Zenburn) as a built-in theme (`theme = "zenburn"`), also usable as a `[custom_theme]` base. 

Encouraged to submit after an online interaction with @benvinegar showing my personal config to use the theme.

```
A warm, low-contrast dark palette — by @ramin, inspired by and slightly modified from the original [Zenburn (https://github.com/jnurmine/Zenburn) by Jani Nurminen. Warm off-white text, cyan-blue functions and types, sage comments, dusty-red strings.
```

  ## Changes

  - New `src/ui/themes/zenburn.ts`, following the existing single-theme pattern (ember/graphite).
  - Registered in `THEMES` and `BUILT_IN_THEME_IDS` (works as a custom-theme `base`).
  - Added to the `hunkdiff/opentui` theme-name union and the component docs.
  - README + CHANGELOG updated; theme/config/opentui tests extended for the new id and label.

  ## Validation

  `typecheck`, `oxlint`, and `oxfmt --check` clean; theme, config, and opentui suites pass.

  ## Screenshot

zenburn

Pairs with. the fix from https://github.com/modem-dev/hunk/pull/393 to allow the theme to be more accurate across keywords

## Follow ups

- personally would be inclined to alphabet sort the themes in some of this, and organize some imports, but wanted to scope this to JUST the theme as much as possible. Would be happy to follow up with some sorting if welcome

Thank you!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable user-visible changes to Hunk are documented in this file.
 ### Added
 
 - Show the newly selected theme in the footer status bar when switching themes.
+- Added a Zenburn built-in theme (`theme = "zenburn"`), a warm low-contrast dark palette inspired by Jani Nurminen's original Zenburn. It also works as a custom-theme `base`.
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ You can persist preferences to a config file:
 Example:
 
 ```toml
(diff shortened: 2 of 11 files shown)


benvinegar:
Looks great.

PR #411, opened 2026-06-10
fix(ui): keep popups opaque in transparent mode

reobin:
:wave: (hope these are welcome)

* keep menu dropdowns and the help dialog on the base theme so transparent background only applies to review/app surfaces
* add interaction coverage for transparent mode to verify menu and help backgrounds stay readable

### DIFF
--- a/src/ui/App.tsx
+++ b/src/ui/App.tsx
@@ -954,7 +954,7 @@ export function App({
             activeMenuSpec={activeMenuSpec}
             activeMenuWidth={activeMenuWidth}
             terminalWidth={terminal.width}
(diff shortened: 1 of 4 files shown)


franky47:
Tried this locally, it works well (before/after):
<table>
<tr>
 <td>
image
</td>
 <td>
CleanShot 2026-06-12 at 10 39 40@2x
</td>
</table>

CI failing is formatting-related, running `bun run format` fixed it.

benvinegar:
@reobin Thanks for submitting, and @franky47 thanks for helping me test! 🙏

PR #418, opened 2026-06-11
feat(ui): honor transparent background in static pager output

franky47:
When using hunk as a pager in lazygit in a transparent-background terminal (Ghostty), the static ANSI output of the pager caused solid backgrounds on context lines and other non-diff elements.

This change uses a similar system as the TUI to mark surfaces as transparent in the pager output, but keeps +/- background tints painted so they're easier to see visually in diffs:

## Before / After

<table>
<tr>
 <td>
 CleanShot 2026-06-11 at 22 15 08@2x
 <td>
 <td>
 CleanShot 2026-06-11 at 22 16 25@2x
 </td>
</table>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


benvinegar (approved):

benvinegar (approved):

benvinegar:
@franky47 Thank you – this was an A+ contribution. The screenshots are a huge help.

PR #422, opened 2026-06-12
perf(ui): yield scroll updates and highlight work to input timers

fink-andreas:
# Scroll Performance Improvements

## Summary

This change makes Hunk scrolling smoother by reducing how much React and background work runs for each scroll tick. The biggest improvement is visible when holding the down arrow key or continuously scrolling with the mouse wheel.

### 1. Batched scroll-position updates

`src/ui/components/panes/DiffPane.tsx`

Previously, every scroll change scheduled a `queueMicrotask` to mirror the OpenTUI scrollbox position into React state. During rapid mouse wheel or key-repeat scrolling, that could turn many tiny native scroll deltas into many React state updates and review-stream re-renders.

Now viewport reads are coalesced with a short frame timer:

```ts
scheduledViewportRead = setTimeout(() => {
  readViewport();
}, 16);
```

In simple terms: instead of doing React work for every scroll event, Hunk updates React at most about once per frame while scroll input is arriving.

### 2. Syntax highlighting yields to input/render timers

`src/ui/diff/pierre.ts`

Previously, serialized syntax highlighting work used `queueMicrotask`:

```ts
queueMicrotask(() => {
  resolve(run());
});
```

Highlighting is CPU-heavy background work. Microtasks run before timers, input, and render callbacks, so a long queue of highlight jobs could delay visible scrolling.

Now highlighting jobs use timer scheduling:

```ts
setTimeout(() => {
  resolve(run());
}, 0);
```

In simple terms: highlighting still runs in the background, but it yields between files so OpenTUI input and frame rendering get more chances to run.

### 3. Hover/add-note clearing is targeted

`src/ui/components/panes/DiffPane.tsx`

Previously, scrolling cleared hover/add-note affordances globally. The clear signal was passed to every visible file section, which could invalidate multiple memoized `DiffSection` renders.

Now Hunk tracks the currently hover-owned file:

```ts
const hoveredFileIdRef = useRef<string | null>(null);
```

On scroll, Hunk only clears hover state when a file actually owns hover actions, and only sends the clear signal to that file:

```tsx
hoverClearSignal={
  addNoteHoverClearFileId === file.id ? addNoteHoverClearSignal : 0
}
```

In simple terms: scrolling no longer tells every visible diff section to update just to hide one hover control.

### 4. Large-stream benchmark measures render cost more directly

`benchmarks/large-stream.ts`

The benchmark previously wrapped scroll/render work in React test `act()` and `Bun.sleep(0)`, which exaggerated scroll costs because it measured test-harness scheduling behavior.

The benchmark now avoids that artificial pattern and measures the OpenTUI render loop more directly. This did not directly fix production scrolling, but helped separate benchmark artifacts from real app scroll-path costs.

## Verification

Commands run:

```powershell
bun run typecheck
bun run lint -- src/ui/components/panes/DiffPane.tsx src/ui/diff/pierre.ts benchmarks/large-stream.ts
bun test src/ui/AppHost.scroll-regression.test.tsx
bun run bench:large-stream
```

Manual testing confirmed smoother continuous mouse-wheel scrolling and smoother held down-arrow scrolling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable user-visible changes to Hunk are documented in this file.
 
 ### Changed
 
+- Coalesced scroll-position React updates into a single per-frame read and shifted background syntax highlighting from microtasks to timers, so rapid wheel or held-arrow scrolling no longer produces visible jank from per-delta state updates or per-file highlight work starving input and render callbacks.
+
 ### Fixed
 
 ## [0.15.1] - 2026-06-09
--- a/benchmarks/large-stream.ts
+++ b/benchmarks/large-stream.ts
@@ -3,6 +3,7 @@ import { performance } from "perf_hooks";
 import React from "react";
 import { testRender } from "@opentui/react/test-utils";
 import { AppHost } from "../src/ui/AppHost";
+import { VIEWPORT_READ_COALESCE_MS } from "../src/ui/lib/viewportTiming";
 import {
   createLargeSplitStreamBootstrap,
   DEFAULT_FILE_COUNT,
@@ -14,7 +15,6 @@ const VIEWPORT = {
   height: 28,
 } as const;
 const SCROLL_TICKS = 4;
-const VIEWPORT_READ_COALESCE_MS = 16;
 const SCROLL_TARGET = {
   x: 170,
   y: 12,
--- a/benchmarks/large-stream.ts
+++ b/benchmarks/large-stream.ts
@@ -14,6 +14,7 @@ const VIEWPORT = {
   height: 28,
 } as const;
 const SCROLL_TICKS = 4;
+const VIEWPORT_READ_COALESCE_MS = 16;
 const SCROLL_TARGET = {
   x: 170,
   y: 12,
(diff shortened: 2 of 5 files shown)


fink-andreas:
Fixed in e991b25: the scroll benchmark now waits for `VIEWPORT_READ_COALESCE_MS + 1` after each wheel event, so the timer-coalesced viewport read updates React state before `renderOnce()` is measured.

Fixed-harness results:

| revision | cold_first_frame_ms | warm_first_frame_ms | windowed_scroll_ticks_ms |
| --- | ---: | ---: | ---: |
| before scroll-render-defer (834f7a8, fixed harness copied over) | 76.76 | 63.22 | 7220.48 |
| after scroll-render-defer (e991b25) | 42.94 | 55.19 | 5802.51 |

So with the viewport-read path included, the scroll metric still improves from 7220.48ms to 5802.51ms for 4 scroll ticks (~20% faster).

benvinegar:
@fink-andreas Thanks for submitting. I made one tweak that removed a magic number (`16`) in favor of a declared const (maybe a bit inelegant having a dedicated file for this but hey, at least discoverable).

PR #442, opened 2026-06-15
test: add unit coverage for untested core and UI logic modules

benvinegar:
## What

Adds colocated unit tests for pure parsing/heuristic modules that previously had **no direct coverage** — the cheapest, most deterministic, and most bug-prone gaps in the suite. 53 new tests across 6 files.

| Module | Tests | Why it was risky |
|---|---|---|
| `core/patch/gitFormat.ts` | 10 | Densest untested logic in the repo: noprefix/mnemonic/quoted/rename header rewriting + the invariant that hunk-body lines mimicking headers stay verbatim |
| `core/patch/normalize.ts` | 12 | Terminal-control stripping (SGR/OSC/DCS), CRLF, path-escape ordering, and the composed `normalizePatchText` pipeline incl. git-log metadata |
| `core/binary.ts` | 16 | `patchLooksBinary` marker matching + the `isProbablyBinaryFile` control-byte heuristic (NUL, 30% threshold, empty, missing file) |
| `ui/lib/responsive.ts` | 8 | The architecturally-central `auto`/split/stack + sidebar selection, including bucket boundaries (159/160/219/220) |
| `ui/lib/keyboard.ts::isSaveDraftNoteKey` | 1 | The one keyboard matcher with no coverage — multi-encoding Ctrl-S (raw / Kitty CSI-u / tmux) |
| `core/diffFile.ts` | 7 | `buildDiffFile` id format, binary fallback, stats/override precedence, fetcher context; `countDiffStats`; skipped-large placeholder |

Each module's actual behavior was probed before writing assertions, so tests pin real behavior rather than assumptions. Binary fixtures use `os.tmpdir()` + node `path` and avoid Windows-reserved filenames for cross-platform safety.

## Verification

- `bun run typecheck` — clean
- `bun run lint` — 0/0
- `bun run format:check` — clean
- Full unit suite — 840 pass, 0 fail (these tests add 53 more)

Reviewed by a Sonnet subagent; all findings (a Windows-reserved filename, a threshold-naming nit, and three coverage/diagnostic improvements) applied.

Test-only; no user-visible change (empty changeset included).

### DIFF
--- a/.changeset/patch-and-core-unit-coverage.md
+++ b/.changeset/patch-and-core-unit-coverage.md
@@ -1,4 +1,4 @@
 ---
 ---
 
-Add unit coverage for previously untested pure-logic modules: git patch prefix normalization, patch text normalization, binary detection, responsive layout selection, the save-draft-note key matcher, and the diff-file model builder. Test-only; no user-visible change.
+Add unit coverage for previously untested pure-logic modules: git patch prefix normalization, patch text normalization, binary detection, responsive layout selection, the save-draft-note key matcher, and the diff-file model builder. Also anchors the `GIT binary patch` marker to start-of-line so it is detected symmetrically with the `Binary files ... differ` marker. No user-visible change.
--- a/.changeset/patch-and-core-unit-coverage.md
+++ b/.changeset/patch-and-core-unit-coverage.md
@@ -1,0 +1,4 @@
+---
+---
+
+Add unit coverage for previously untested pure-logic modules: git patch prefix normalization, patch text normalization, binary detection, responsive layout selection, the save-draft-note key matcher, and the diff-file model builder. Test-only; no user-visible change.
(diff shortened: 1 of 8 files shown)


benvinegar:
Thanks @greptile-apps — both addressed in 0cdcbfe:

**Issue 1 (GIT binary patch marker inconsistency):** Took the preferred fix rather than pinning the limitation. `patchLooksBinary` now anchors the `GIT binary patch` marker with `(^|\n)...(?:\n|$)`, symmetric with the `Binary files ... differ` marker, so it's detected even as the first block in a patch. The test now asserts `true` for the start-of-string case. No real-world behavior change (a git binary patch is always preceded by its `diff --git`/`index` header), so this stays a no-user-visible-change entry.

**Issue 2 (`as Partial<KeyEvent>` cast):** The cast was actually redundant — `raw: string` is a declared field on `KeyEvent` (`ParsedKey`), which is why `keyboard.ts` reads `key.raw` without a cast. Removed both casts so `raw` is passed directly; this *restores* the type-level signal you wanted — if upstream ever drops `raw`, these lines now fail to compile instead of silently passing.

PR #445, opened 2026-06-16
test: cover the VCS adapter layer

benvinegar:
## What

Lifts `src/core/vcs` coverage from **81% → 90%**.

| File | Before | After |
|---|---|---|
| `index.ts` | 92% | **100%** |
| `git.ts` | 94% | **~99%** |
| `jj.ts` | 84% | **89%** |
| `sl.ts` | 36% | **56%** |

### `index.ts`
Adds the unknown-VCS-id throw and the non-stash `createUnsupportedVcsOperationError` message.

### `git.ts`
Exports the pure helpers (`parseGitNumstat`, `shouldSkipLargeTrackedDiff`, `gitEndpointSourceSpec`, `statSignature`) and unit-tests their edge cases — malformed numstat entries, the line/byte large-file budgets, the missing-file `statSync` fallback, every diff-endpoint source-spec kind, and present-vs-missing stat signatures. Plus real-git `watchSignature` coverage (working-tree / revision / stash) and a walk-to-root null-detect case.

### `jj.ts` / `sl.ts`
Adds **binary-free** tests for the branches that run before any `jj`/`sl` invocation: nested `.jj`/`.sl` detection, walk-to-root null, staged-diff rejection, and stash-show rejection in both `loadReview` and `watchSignature`. `sl` also covers the `.hg`-without-`requires` fallback (a `.hg` dir that isn't Sapling).

## Why `sl` is still 56%
The bulk of `sl.ts` (and the rest of `jj.ts`) is the **diff/show exec glue** that shells out to the `sl`/`jj` binaries. Those paths are covered by the existing `test.skipIf(!available)` integration tests on machines where the VCS is installed. Driving them in CI would require either module-mocking (which **leaks across files** in Bun — I verified capture-and-restore doesn't contain it) or a fake on-PATH binary (fragile + non-portable), so I left them as binary-gated integration coverage rather than add brittle tests for thin pass-through glue.

## Production change
The only non-test change is exporting four previously-internal pure helpers in `git.ts`, marked "Exported for unit testing."

## Verification
- `bun run typecheck`, `bun run lint`, `bun run format:check` — all clean
- Full unit + CLI + session suite — **861 pass, 0 fail**

Test-only; empty changeset included. No user-visible change.

### DIFF
--- a/.changeset/vcs-adapter-coverage.md
+++ b/.changeset/vcs-adapter-coverage.md
@@ -1,0 +1,4 @@
+---
+---
+
+Add unit coverage for the VCS adapter layer: registry error paths, Git numstat parsing, large-file skip heuristics, diff-endpoint source-spec mapping, stat signatures, and Git watch signatures, plus binary-free detection and unsupported-operation branches for the Jujutsu and Sapling adapters. Several pure Git helpers are now exported for direct testing. Test-only; no user-visible change.
--- a/src/core/vcs/git.test.ts
+++ b/src/core/vcs/git.test.ts
@@ -2,7 +2,13 @@ import { afterEach, describe, expect, test } from "bun:test";
 import { mkdirSync, mkdtempSync, realpathSync, rmSync, writeFileSync } from "node:fs";
 import { platform, tmpdir } from "node:os";
 import { join } from "node:path";
-import { gitAdapter } from "./git";
+import {
+  gitAdapter,
+  gitEndpointSourceSpec,
+  parseGitNumstat,
+  shouldSkipLargeTrackedDiff,
+  statSignature,
+} from "./git";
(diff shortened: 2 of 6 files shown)


benvinegar:
Thanks @greptile-apps — both fixed in b33d97b:

1. **Vacuous diff signature:** Reordered so the `working-tree-diff` signature is measured while the tree is still dirty (before the stash push). It now asserts the signature contains the tracked `diff --git a/file.txt b/file.txt` and an `untracked:` stat fragment — which also gives real coverage of `statSignature`'s present-file branch. The stash is created afterward for the stash-show case.
2. **Binary numstat:** Added an explicit case asserting `-\t-\tsrc/logo.png` is dropped (non-numeric counts fail the finite guard) while a sibling numeric entry survives, pinning the silent-drop behavior.

PR #447, opened 2026-06-16
feat: high-contrast themes

edemaine:
I have a hard time seeing the current themes. Here are some high-contrast themes that seem easier to read (though further tweaking is certainly possible). Hopefully this is helpful for accessibility in general (not just me 🙂).

I wasn't sure how to order the themes, so I put the new ones last. Let me know if you prefer a different order.

## Screenshots

Comparison against some representative existing dark/light themes:

### paper (existing light)
paper

### high-contrast light (NEW)
high-contrast-light

---

### graphite (existing dark)
graphite

### high-contrast dark (NEW)
high-contrast-dark


benvinegar:
For the record I'm in the same boat: I think the current themes are too low contrast.

But I don't know that offering a completely different set of themes is right either.

Let me think on this.

edemaine:
FWIW, VScode offers five separate high-contrast themes: Dark High Contrast, Light High Contrast, GitHub Dark High Constrast, GitHub Light High Contrast, and Monokai High Contrast. (I use Dark High Contrast personally.) GitHub similarly offers high-contrast themes which is what two of those are based on, and Shiki copies these GitHub themes.

I think my theme was a bit higher contrast (similar to VSCode's I think) and I prefer it... but the new github-dark-high-contrast theme from Shiki is probably good enough, and just using Shiki themes is probably easier, so I'm not going to pursue this further. Although I guess I could try a PR to Shiki...?

## Shiki github-dark-high-contrast
high-contrast-range-github-dark-high-contrast

### high-contrast dark from this PR
high-contrast-dark

PR #452, opened 2026-06-18
fix(build): prevent caller bunfig autoload

jrpat:
Add Bun's compile-time autoload disabling flag to the standalone binary build paths so distributed Hunk executables do not load bunfig.toml from the directory they are launched in.

The change is scoped to bunfig.toml autoloading - I did not include `--no-compile-autoload-dotenv`, which is a similar behavior but may not actually be preferable.


benvinegar:
@jrpat Thanks! I hit this once and didn't know how to fix it.

Issue #113, opened 2026-03-24
Persist live comments so a Hunk review session can be continued after quit


benvinegar:
## Problem

If you quit Hunk, you lose your live agent comments for that review session.

That makes longer review workflows fragile: comments are useful precisely because they accumulate context, but today that context appears to be tied to the lifetime of the live session.

## Current implementation notes

Relevant pieces:
- #53 — `feat: add live session comment lifecycle`
- `src/mcp/sessionRegistration.ts`
  - `createInitialSessionSnapshot()` starts each session with `liveComments: []`
- `src/mcp/daemonState.ts`
  - session snapshots are tracked in daemon memory while the session is connected

That is great for live coordination, but there does not appear to be a persistence / restore path for comments or session state after the TUI exits.

## Desired behavior

There should be a way to continue a previous Hunk review session and restore its live comments (and ideally related state like the focused file/hunk).

The UX could be something like:
- `hunk --continue-session`
- `hunk session resume ...`
- or another explicit restore flow

## Scope suggestions

Persist at least:
- review input identity (working tree / show ref / patch file / etc.)
- live comments
- selected file / selected hunk
- maybe note visibility / other lightweight view state

A local-only sidecar under `.hunk/` seems like a reasonable starting point.

## Acceptance criteria

- A user can quit Hunk and later reopen the same review with comments restored.
- Restored sessions reconnect to the same review input, not just the same repo.
- Restoring a session also restores the current focus when possible.
- There is an explicit way to discard stale/old saved sessions.
- Add coverage for saving and restoring at least one live-comment session.

## Notes

This issue is about persistence of review context, not only daemon uptime. Even if the daemon exits cleanly, users should still be able to resume a meaningful review later.

frederikschubertflex:
Thank you first of all for this nice tool! I am using it in my workflow but I started to use git notes to persist the comments. I just wanted to leave that feedback here before I put up the effort and create a PR (prompt request as many call it now 😆 ).

heyitsaamir:
Related, if there are comments in a diff session, it'd be great if there was friction in quitting the session so an accidental escape doesn't cause you to lose all your comments

benvinegar:
@heyitsaamir Good idea!

tolgaerdonmez:
[...]

PR #473, opened 2026-06-22
persist human review notes to a JSON sidecar via --store-notes

muratovv:
## What
Opt-in `--store-notes <path>` (cwd-relative) persists human `c` review notes to a JSON sidecar keyed by file id, readable directly off disk.

## Why
Human notes currently die with the session, so an agent can't pick up your review after you quit — the reverse of `--agent-context`. Addresses #113 (persist notes past quit; suggests a `.hunk/` sidecar).

## Behavior
Opt-in; default behavior unchanged. Disk I/O is best-effort and never crashes the UI; startup warning if the path isn't writable. Includes tests, `docs/agent-workflows.md`, and a `minor` changeset.

### DIFF
--- a/.changeset/persist-review-notes.md
+++ b/.changeset/persist-review-notes.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": minor
+---
+
+Add an opt-in `--store-notes <path>` flag that persists human review notes to a JSON sidecar (cwd-relative). Notes survive closing the TUI and can be read back off disk by an agent. Omitting the flag keeps notes in-memory only, preserving current behavior.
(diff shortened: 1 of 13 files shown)


muratovv:
Hey @benvinegar 👋 gentle nudge on this one whenever you get a chance — I've addressed the earlier review feedback, so it should be in decent shape now, but honestly what I'd most like to know is whether a feature like this is something you'd actually want in hunk at all. If it doesn't feel right the way it's currently shaped, I'm completely happy to rework it — just tell me what you'd want it to look like and I'll take it from there. No rush at all, and thanks for hunk!

PR #480, opened 2026-06-23
bugfix: Let draft notes own keys in pager mode

skatkov:
When Hunk was launched as a pager from git diff, typing in a draft inline note could trigger shortcuts. As example - instead of inserting `s` into the note, Hunk toggled the sidebar because pager-mode global shortcuts were handled before the focused note input.

## Steps
Here is a video of this problem

https://github.com/user-attachments/assets/65cc5a8f-5042-4d4e-97ba-0a49d1661756

1. open hunk pager through `git diff`
2. Start an inline note
3. Try to write `test`

After typing `s` sidebar is opened, focus from the inline note is lost, and I can't type anymore into it.

## Solution
- Fixes pager-mode draft note input so typed characters are not intercepted as pager shortcuts.
- Moves focused-input keyboard handling before pager shortcut handling in `useAppKeyboardShortcuts`.
- Adds a PTY regression test covering a draft note opened in `patch --pager`, verifying `s` is inserted into the note instead of toggling the sidebar.

### DIFF
--- a/src/ui/hooks/useAppKeyboardShortcuts.ts
+++ b/src/ui/hooks/useAppKeyboardShortcuts.ts
@@ -570,11 +570,6 @@ export function useAppKeyboardShortcuts({
       return;
     }
 
-    if (pagerModeRef.current) {
-      handlePagerShortcut(key);
-      return;
-    }
-
     if (handleDialogShortcut(key)) {
       return;
     }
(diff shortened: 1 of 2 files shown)


skatkov:
@benvinegar sorry for pinging.

but this is quite an annoying bug, that would help me tremendously to make hunk usable for my workflow. 

can you please check this out? 

Thank you.

PR #483, opened 2026-06-24
feat(ui): add PR-style changeset overview with title and description

borgand:
Agents can now set top-level `title` and `description` fields in the --agent-context JSON; Hunk displays them in a toggleable Overview overlay (`o`) that auto-opens once when a title or description is present.

Includes a pure markdown-to-rows renderer with heading level differentiation, wrapText moved to a shared text helper, and updated docs covering the agent-context JSON format.

The intended use case is that agent will conclude longer work session with updating the hunk session with the overview, explaining the whole changeset as people would do for PR-s

### DIFF
--- a/.changeset/pr-style-changeset-description.md
+++ b/.changeset/pr-style-changeset-description.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": minor
+---
+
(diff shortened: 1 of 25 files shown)


benvinegar:
This is a non-trivial patch (~1000 LoC) to add this. Could you help me understand what it looks like before/after? e.g. screenshots or video etc

borgand:
Yes, I'm AFK mostly currently, but let me try to update this with some images.

PR #485, opened 2026-06-26
feat: highlight .mts and .cts files as TypeScript

scaryrawr:
[Node has native support for TypeScript](https://nodejs.org/learn/typescript/run-natively), but it treats `.ts` files differently based on if your package.json type is ESM/CJS.

To get consistent behavior they use `.cts` or `.mts` so the script is treated as CJS/ESM even if it does not line up with the package type.

pierre (and hunk) don't support these extensions by default for syntax highlighting:
image

with the custom extensions we get the expected syntax highlighting:
image

There's also [a PR to pierre](https://github.com/pierrecomputer/pierre/pull/893) for adding `mts` and `cts` to their file type map so this will hopefully not be needed in the long term.

### DIFF
--- a/.changeset/mts-cts-typescript-highlighting.md
+++ b/.changeset/mts-cts-typescript-highlighting.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Highlight `.mts` and `.cts` files as TypeScript instead of plain text.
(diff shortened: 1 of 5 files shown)


benvinegar:
This solution seems easy/clean - done!

PR #486, opened 2026-06-26
feat(watch): live-reload theme and chrome on config change

eduwass:
## Summary

When running `hunk diff --watch`, the viewer now live-reloads its theme and chrome the moment Hunk's config changes — no restart required. Previously `--watch` only tracked the diff input, so editing `~/.config/hunk/config.toml` (e.g. switching a `[custom_theme]` palette from a theming tool) had no effect until you quit and relaunched.

## What this adds

The watch loop now also watches Hunk's config files and reloads on change:

- The global config (`~/.config/hunk/config.toml`) and the repo-local config (`.hunk/config.toml`) are folded into the existing diff-input watch poll via a new `computeConfigSignature()` in `core/watch.ts`.
- Keeping it in the **same** poll loop means there's a single in-flight guard, so a config reload can never race the diff reload.
- On any change (input or config) it reuses `refreshCurrentInput()`, which re-runs the normal config resolution pipeline — so freshly-read `[custom_theme]` / `[custom_theme.syntax]` palettes repaint live — while preserving the active view options and any CLI overrides (no `--theme`/`--mode` flag gets silently clobbered by config).

Only active in `--watch` mode, so default one-shot reviews are unchanged.

## Testing

- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run test:theme-contrast`
- `bun test src/core/watch.test.ts` — added unit coverage for `computeConfigSignature` (missing global config, change detection, repo-local `.hunk/config.toml` pickup)
- Real render smoke test: ran `hunk diff --watch` against a repo, switched the active theme so a tool rewrote `[custom_theme]`, and confirmed the running viewer repainted from a dark to a light palette (background `#2D2B55` → `#f9f9f9`, panels and diff bands updated) without a restart, and back again.

### DIFF
--- a/.changeset/live-config-reload.md
+++ b/.changeset/live-config-reload.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": minor
+---
+
+Live-reload the theme and chrome whenever Hunk's config changes — no `--watch` required. The viewer always polls the global (`~/.config/hunk/config.toml`) and repo-local (`.hunk/config.toml`) config files and reloads through the existing reload path on change, the way VS Code / Cursor apply settings on save — so edits to `[custom_theme]`, layout, and other view options take effect without restarting the viewer. In `--watch` mode the same loop additionally tracks the diff input.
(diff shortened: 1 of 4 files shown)


benvinegar:
I wonder if it should just do this always? Vs `--watch`

Precedent seems to be VS Code, Cursor, etc. Those all basically apply their settings immediately, right?

eduwass:
Hey @benvinegar yeah good point. VSCode/Cursor would update automatically, probably what we'd want here too.

eduwass:
@benvinegar made it always-on now, no `--watch` required. The viewer always polls the config files (global + repo-local) and live-reloads theme/chrome on change, the way VS Code / Cursor apply settings on save.

PR #496, opened 2026-06-30
feat: add support line number in "Open file in editor" for helix

iniw:
Does what it says on the tin.

### DIFF
--- a/src/ui/lib/openInEditor.ts
+++ b/src/ui/lib/openInEditor.ts
@@ -68,6 +68,10 @@ export function buildEditorCommand({
     return { command, args: [...editorArgs, "--goto", `${filePath}:${line}`] };
   }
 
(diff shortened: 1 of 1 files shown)


benvinegar:
Is there a way to make this editor agnostic somehow? Don't love "if app == hx" in the code (would set a new precedent).

iniw:
> Don't love "if app == hx" in the code (would set a new precedent).

I don't see how it's different/better/worse than the other `includes()` that preceded it.

> Is there a way to make this editor agnostic somehow?

Don't think so. The editors present different CLI interfaces so we just have to deal with them 🤷

benvinegar:
> I don't see how it's different/better/worse than the other includes() that preceded it.

Okay, re-reading the code again, you're right.

PR #497, opened 2026-06-30
Update hunk installation instructions in README

fraluc06:
Update brew installation method. @benvinegar


benvinegar:
@fraluc06 Thanks!

PR #498, opened 2026-06-30
Recommend installation from the main homebrew tap

lugray:
It is now available there, is more up to date than the custom tap, and doesn't need to be trusted with homebrew's new trust system.


benvinegar:
@lugray Thank you! 🇨🇦 🫡

PR #500, opened 2026-06-30
feat(session): filter comment list by author

### LINKED ISSUE #489: Add --author filter to hunk session comment list

Add an `--author <name>` filter to `hunk session comment list` so the CLI can return only comments written by one specific author.

## Why?

The `comment add` and `comment apply` subcommands already let callers tag a comment with `--author <name>` (or the `author` field in the stdin JSON for `apply`), and `comment list` already supports `--file <path>` and `--type <live|all|ai|agent|user>` filters. The one missing dimension is author.

In a typical agent-assisted review a coding agent and a human both leave notes on the same session, and the JSON output of `hunk session comment list --json` ends up mixed. Right now the only way to pull out one author's notes is to pipe the JSON through `jq`:

```
hunk session comment list --json | jq '.[] | select(.author=="pi")'
```

That works, but it pushes what is conceptually a one-flag filter into every consumer that needs it. A native `--author <name>` flag mirrors the same ergonomics as `--file` and `--type` and keeps `--json` consumers thin.

## How? (optional)

- Add `.option("--author <name>", "filter comments to one author")` to the `session comment list` commander setup, parse it like the other string filters, and forward it through `SessionCommentListCommandInput`.
- In the daemon/list handler, drop comments whose `author` does not match the flag (case-sensitive, exact match — same shape as `--type`).
- Treat a missing `author` field as a separate value: add a companion `--no-author` (or `--author=""`) flag that returns only comments without an `author`. Two flags keeps the matcher unambiguous and mirrors how `comment clear` already uses `--include-user` / `--all`.
- Add unit coverage next to `src/core/cli.test.ts` for the new flag, plus a daemon-level test that the returned list honors the filter.
- Update `docs/agent-workflows.md` (or the equivalent CLI reference) with one example: `hunk session comment list --author pi --json`.

abgregs:
## What

Adds `--author <name>` and `--no-author` filters to `hunk session comment list`, letting the CLI narrow inline review comments by author — mirroring the existing `--file` and `--type` filters.

## Why

In an agent-assisted review, an agent and a human both leave notes on the same session, so `hunk session comment list --json` returns a mixed set. Today the only way to isolate one author's notes is to pipe the JSON through `jq` (`jq '[.comments[] | select(.author=="pi")]'`), which pushes a conceptually one-flag filter into every consumer. A native `--author <name>` flag — plus a `--no-author` companion for untagged notes — keeps `--json` consumers thin and matches the ergonomics of `--file` / `--type`.

## How it works

The filter threads through the same layers as `--type`, with the actual filtering in a single place (the daemon list handler):

- **`core/cli.ts`** — `--author <name>` / `--no-author` options, parsed and forwarded into the command input
- **`core/types.ts`, `session/protocol.ts`** — `author` / `noAuthor` on the command input and wire shapes
- **`hunk-session/cli.ts`** — the client includes them in the request
- **`session-broker/brokerServer.ts`** — one post-branch filter narrows the result by author, covering **both** the live-comment and review-note paths (both carry `author`)

`--author <name>` matches that author exactly; `--no-author` returns only comments with **no author tag** — comments added via `comment add`/`apply` without an author. (Human `c` notes are auto-tagged `author: "user"`, so they're matched by `--author user`.) Neither flag returns everything. Documented in `docs/agent-workflows.md`.

## Testing

- [x] **Parse** (`core/cli.test.ts`): `--author pi` → `{ author: "pi" }`; `--no-author` → `{ noAuthor: true }`.
- [x] **Filter** (`brokerServer.helpers.test.ts`): for a mixed-author comment set, `--author` returns only the matching author, `--no-author` returns only the author-less, and no flag returns all.
- [x] `bun run typecheck`, `bun run lint`, and `bun run format:check` clean.
- [x] Verified end-to-end on macOS (screenshots below).

## Screenshots

End-to-end via `hunkdev` on this branch — drag each PNG into the GitHub editor to upload it in place:

**1. Add one `--author pi` comment and one untagged comment**

add-authored-and-authorless-comments

**2. `hunk session comment list --author pi` → only pi's**

list-pi-authored

**3. `hunk session comment list --no-author` → only the untagged one**

list-authorless

**4. `hunk session comment list` (no filter) → both**

list-all

Fixes #489

### DIFF
--- a/.changeset/comment-list-author-filter.md
+++ b/.changeset/comment-list-author-filter.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": minor
+---
+
+Add `--author <name>` and `--no-author` filters to `hunk session comment list` to narrow inline review comments by author.
--- a/docs/agent-workflows.md
+++ b/docs/agent-workflows.md
@@ -106,6 +106,8 @@ For comment cleanup and inspection, use:
 
 ```bash
 hunk session comment list --repo .
+hunk session comment list --repo . --author pi --json # only comments authored by "pi"
+hunk session comment list --repo . --no-author --json # only comments with no author tag
 hunk session comment rm --repo . <comment-id>
 hunk session comment clear --repo . --file README.md --yes
 hunk session comment clear --repo . --all --yes # also clears human `c` notes
--- a/src/core/cli.test.ts
+++ b/src/core/cli.test.ts
@@ -715,6 +715,47 @@ describe("parseCli", () => {
     ).rejects.toThrow("Comment type must be one of live, all, ai, agent, or user.");
   });
 
+  test("parses session comment list with an author filter", async () => {
+    const parsed = await parseCli([
+      "bun",
+      "hunk",
+      "session",
+      "comment",
+      "list",
+      "session-1",
+      "--author",
+      "pi",
+    ]);
+
+    expect(parsed).toEqual({
+      kind: "session",
+      action: "comment-list",
+      selector: { sessionId: "session-1" },
+      author: "pi",
+      output: "text",
+    });
+  });
+
+  test("parses session comment list with the no-author filter", async () => {
+    const parsed = await parseCli([
+      "bun",
+      "hunk",
+      "session",
+      "comment",
+      "list",
+      "session-1",
+      "--no-author",
+    ]);
+
+    expect(parsed).toEqual({
+      kind: "session",
+      action: "comment-list",
+      selector: { sessionId: "session-1" },
+      noAuthor: true,
+      output: "text",
+    });
+  });
+
   test("parses session comment rm", async () => {
     const parsed = await parseCli([
       "bun",
--- a/src/core/cli.ts
+++ b/src/core/cli.ts
@@ -902,7 +902,7 @@ async function parseSessionCommand(tokens: string[]): Promise<ParsedCliInput> {
             "Usage:",
             "  hunk session comment add (<session-id> | --repo <path>) --file <path> (--old-line <n> | --new-line <n>) --summary <text> [--focus]",
             "  hunk session comment apply (<session-id> | --repo <path>) --stdin [--focus]",
-            "  hunk session comment list (<session-id> | --repo <path>) [--file <path>] [--type <live|all|ai|agent|user>]",
+            "  hunk session comment list (<session-id> | --repo <path>) [--file <path>] [--type <live|all|ai|agent|user>] [--author <name>] [--no-author]",
             "  hunk session comment rm (<session-id> | --repo <path>) <comment-id>",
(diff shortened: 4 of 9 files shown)


benvinegar:
Don't love --no-author meaning "human notes", but I guess we can go with it. Wish there was better / clearer language ... I feel agents would perform better.

abgregs:
> Don't love `--no-author` meaning "human notes" ... Wish there was better / clearer language

Yeah agreed the language is a bit confusing, and I just realized my PR description was wrong about who actually lands in `--no-author` (details below). On the naming itself, I stuck with what #489 proposed, but my original description said `--no-author` "returns only comments with no author tag (human `c` notes, or agent comments added without `--author`)" mistakenly implying human `c` notes land in `--no-author`.

For context on `--no-author`, #489 asked to "treat a missing `author` field as a separate value" via a dedicated flag ("two flags keeps the matcher unambiguous"). Re-tracing the code, `--type` filters by writer category (the `source` field — `ai` / `agent` / `user`), while `--author` filters the optional identity label within a category. So `--no-author` means only "no identity label" — nothing about human vs agent. Human `c` notes are actually auto-tagged `author: "user"`, so `--author user` matches them; the `--no-author` set is just comments without an author tag.

I've corrected the PR description accordingly so `--no-author` now reads "no author tag," which holds regardless of where we land on naming. Let me know if that tracks.

Sticking with the `--no-author` naming, I could work in your idea to improve the agent-facing language by sharpening it in these places:

- help text — becomes "only comments with no author tag"
- both agent docs — `docs/agent-workflows.md` and `skills/hunk-review/SKILL.md` get the same one-sentence model: `--type` = writer category (`source`), `--author` = identity tag within it; human `c` notes are tagged `user`, so use `--author user` for those
- the skill's command synopsis — noticed `SKILL.md` still lists `comment list` with only `--file`/`--type`, so an agent following the recommended workflow never learns the new flags exist; I'd add them there

Happy to give that a shot if it sounds right to you.

abgregs:
Rebased onto main (ported to the new packages/ layout and the agent command registry) and applied the suggested wording changes from above as a separate commit:

- help text: `--author <name>` → "only comments tagged with this author", `--no-author` → "only comments with no author tag"
- `docs/agent-workflows.md` and the generated hunk-review skill now explain `--type` (writer category) vs `--author` (tag within it)
- since flags now live in `surface.ts`, the skill's comment list synopsis picks up `--author `/ `--no-author `automatically

Verified live: human c notes are tagged `author: "user"`, so `--author` user matches them, while `--no-author` returns only comments with no author tag:

Screenshot 2026-09-16 at 5 00 13 PM

Also updated the PR description to match the new layout. Ready for another look when you have a minute.

Regarding the failed checks besides the Vercel deploy the PTY scrolling test timed out on the latest run; it passed on the previous run with identical code (the only change since is regenerated website docs), so it looks flaky.

PR #519, opened 2026-07-09
fix(ui): distinguish root files in sidebar

### LINKED ISSUE #461: Visually distinguish files of different levels in the sidebar
### What happened?

  When a file of upper level appears after files from a subfolder, the sidebar can make the file look like it belongs to the preceding folder.

### Steps to reproduce

For example, given this file order:

- `subfolder/file.txt`
- `zzz-root.txt`

the sidebar is approximately:

```text
subfolder/
  D file.txt
  M zzz-root.txt
```

Because root-level files and files inside folder groups use the same
indentation, `zzz-root.txt` appears to be inside `subfolder/`.

This can happen because the sidebar preserves the review stream order and
only renders headers for non-root directories.

### Expected behavior

Root-level files should be visually distinguishable from files belonging to the preceding folder group.

One possible presentation would be to indent grouped files one level further:

```text
subfolder/
    D file.txt
M zzz-root.txt
```

### Version

0.16.0

BowlOfSoup:
## Summary

- Label each run of repository-root files with an in-place `./` sidebar header.
- Preserve the existing VCS or agent-provided review order.
- Keep root files visually distinct from the preceding folder group.

Closes #461

*This PR description was generated by Pi using OpenAI GPT-5.2*


benvinegar:
Thanks for tackling #461 — I reproduced the ambiguity and agree that root files need a clearer boundary.

We tried an alternative that keeps the original VCS or agent-provided review order intact: emit a synthetic `./` group header whenever the ordered sidebar stream enters a run of repo-root files.

```text
nested/
 M file.ts
./
 M zzz-root.ts
```

This makes the root transition explicit without hoisting files or changing the main review stream and navigation order. We have a small working patch with helper and rendered-component coverage.

Since maintainer edits are enabled, I’m happy to push the follow-up onto this branch, or you can adapt the PR yourself if you prefer. Would you be comfortable pivoting #519 to this presentation?

*This comment was generated by Pi using OpenAI GPT-5.2*

BowlOfSoup:
> Since maintainer edits are enabled, I’m happy to push the follow-up onto this branch

Sure, go ahead. A few questions about the alternate proposal:

- Does it mean that indentation is always happening, for all directories and the root `./`? On split screen terminals this might lead to space issues.
- The navigation order: why is it important to keep the VCS order, and not e.g. an alphabetical order?
- The proposed navigation order: does it mean that the list is scattered with `./` if there are lots of root directory changes? Hoisting (and alphabetically, but I didn't to that either) would give a better overview imo.

benvinegar:
> Does it mean that indentation is always happening, for all directories and the root ./? On split screen terminals this might lead to space issues.

The indentation accommodates the (M)odified indicator and the active file indicator. It's always there.

> The navigation order: why is it important to keep the VCS order, and not e.g. an alphabetical order?

Because we're just trying to do the default thing everyone is used to.

> does it mean that the list is scattered with ./ if there are lots of root directory changes?

There would just be one?

benvinegar:
Thank you!

PR #520, opened 2026-07-09
Pause watch polling automatically while sessions are idle

tridha643:
## Summary

- Put every `--watch` session into an automatic low-cost sleep after one minute without keyboard or mouse interaction.
- Stop scheduling Git-backed watch signature polling while asleep, so unattended Hunk tabs do not keep spawning expensive repository checks.
- Wake on the next interaction with one immediate refresh, then resume normal polling.
- Keep wake-up and interval refreshes behind one shared in-flight guard so reactivation cannot start duplicate reloads.
- Avoid adding a new CLI flag, config key, or environment variable.

## Why

It is easy to leave several `hunk diff --watch` tabs open while agents continue changing a repository. Those unattended sessions previously kept polling Git every 250 ms. Automatic sleep removes that background cost without asking users to configure each tab.

Background diff reloads do not count as user activity, so a busy repository cannot keep an unattended session awake forever.

## Testing

- Added interaction coverage for active to idle to active transitions and the immediate wake-up refresh.
- Added coverage proving background reloads do not postpone automatic sleep.
- Kept slow-reload coverage proving wake-up and resumed polling cannot overlap refreshes.
- `bun run typecheck`
- `bun run lint`
- `bun test ./src ./packages ./scripts ./test/cli ./test/session` (1,012 passed, 13 skipped)
- `bun test ./test/pty` (51 passed)
- `HUNK_RUN_TTY_SMOKE=1 bun test ./test/smoke` (9 passed)

## Review guide

- Start with `src/core/watch.ts` for the built-in inactivity threshold.
- Review `src/ui/App.tsx` for activity tracking, sleep/wake transitions, and refresh coalescing.
- Review `src/ui/AppHost.interactions.test.tsx` for the lifecycle and background-reload regressions.

## Tradeoff

A passively viewed watch session stops updating after one minute without input. The next key, click, drag, or scroll performs an immediate refresh before normal polling resumes.


tridha643:
@greptile

tridha643:
@greptile

benvinegar:
Closing as superseded by #531, which replaced the 250 ms Git polling loop with event-driven filesystem observation and infrequent safety checks. The expensive background polling this idle state machine was designed to suppress no longer exists.

PR #521, opened 2026-07-10
fix: reduce retained geometry memory in large reviews

matthew-hre:
Hello! Been a big fan of Hunk recently, I use it basically every day at work -- however, it's been a bit of a memory-hog on my machine, so I thought I'd start doing some digging and hopefully contribute some performance patches! First contribution on this project, let me know if I missed anything.

This PR introduces a fix that reduces memory usage in large reviews by making diff section geometry hold on to less data. 

Previously, `DiffSectionGeometry` was holding on to the full planned row stream for every measured file section -- most review interactions really only need lightweight geometry data: row bounds, hunk bounds, body height, and lookup maps.

Since we already window the rendered rows, keeping full planned rows for the entire changeset means we're still retaining a bunch of row objects, even when most files weren't mounted.

This changes geometry to keep the bounds/navigation data eagerly, but rebuild planned rows lazily, only when copy selection asks for row content.

I also added a focused `bench:geometry-memory` benchmark so this path is easier to measure next time.

## Results

The biggest improvement showed up on the huge synthetic review benchmark:

| Metric | Main | This PR | Difference |
|---|---:|---:|---:|
| first-frame heap | 228.5MB | 175.1MB | -53.4MB / -23% |
| after-navigation heap | 324.4MB | 248.9MB | -75.5MB / -23% |
| after-navigation RSS | 708.8MB | 644.3MB | -64.4MB / -9% |

The resize memory benchmark also improved, which makes sense:

| Metric | Main | This PR | Difference |
|---|---:|---:|---:|
| final heap | 149.7MB | 120.0MB | -29.6MB / -20% |
| heap growth | 110.2MB | 88.0MB | -22.1MB / -20% |
| peak RSS | 481.2MB | 453.1MB | -28.2MB / -6% |
| RSS growth | 216.6MB | 186.8MB | -29.8MB / -14% |

Little less impressive on the `memory.ts` benchmark, but still positive:

| Metric | Main median | This PR median | Difference |
|---|---:|---:|---:|
| first-frame heap | 35.6MB | 31.3MB | -4.3MB / -12% |
| after-navigation RSS | 481.2MB | 465.4MB | -15.8MB / -3% |

I also ran the new focused geometry benchmark on this branch:

```text
bun run benchmarks/geometry-memory.ts
METRIC geometry_heap_growth_bytes=7469892
METRIC geometry_rss_growth_bytes=5898240
METRIC materialize_planned_rows_ms=6.17
METRIC materialized_planned_rows=10260
METRIC materialized_planned_rows_heap_growth_bytes=0
```

This should be fairly cut-and-dry. Rendering still builds planned rows for mounted sections, and copy selection still gets planned rows when it needs them. The only difference really is that cached geometry no longer keeps those rows alive for every file by default.

### DIFF
--- a/.changeset/lazy-geometry-memory.md
+++ b/.changeset/lazy-geometry-memory.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Reduced retained memory for large reviews by lazily materializing cached geometry row plans only when copy selection needs them.
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -62,7 +62,7 @@ bun run bench:competitors
 - `huge-stream.ts` — opt-in huge tier (`--include-huge` or `HUNK_BENCH_INCLUDE_HUGE=1`): cold first frame, scroll-tick and hunk-navigation latency, and memory ceilings on ~1k files / 300k+ diff lines plus one giant ~50k-line file.
 - `large-stream-profile.ts` — optional local profiler for the main pure planning stages behind the large split-stream benchmark.
 - `memory.ts` — optional local RSS/heap profiler after fixture loading, planning, first frame, and next-hunk navigation.
-- `geometry-memory.ts` — optional local retained-memory profiler for all-files section geometry, including the lazy planned-row materialization path used by copy selection.
+- `geometry-memory.ts` — optional local retained-memory profiler for all-files section geometry, including JSC-native heap metrics and giant-file lazy planned-row materialization latency used by first copy selection.
 - `navigation-memory.ts` — optional local retained-memory profiler for repeated hunk navigation through a mounted review stream.
 - `resize-memory.ts` — optional local retained-memory profiler for repeated terminal-width changes through a mounted review stream; this targets geometry-cache retention across resize variants.
 - `competitors.ts` — optional local informational comparisons against `git diff --no-ext-diff`, `delta`, `difftastic`, and `diff-so-fancy` when installed.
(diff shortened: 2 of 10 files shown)


benvinegar:
@matthew-hre Seems reasonable to me, but had GPT 5.6 Sol poke at it from a few angles. It found the following. Thoughts?

```
 1. Medium — the new heap metrics are unreliable
    benchmarks/geometry-memory.ts:90-99,155-165 uses process.memoryUsage().heapUsed. Under the
    documented Bun 1.3.14, all samples reported exactly 39,105,781 bytes, producing false
    zero-growth results while RSS and external memory increased. Use bun:jsc’s
    heapStats().heapSize or omit these metrics.

 2. Medium — first-copy latency is significant but the benchmark misses it
    src/ui/diff/diffSectionGeometry.ts:156-165,378-380 rebuilds the complete row plan when
    copy selection first accesses plannedRows. With the existing 50k-line giant-file fixture,
    I measured 67ms to materialize 44,009 rows. The new benchmark only exercises smaller
    fixed-size diffs and reports ~2–6ms. It should cover the giant-file path or materialize
    only the selected range.

 3. Low — lazy rows can diverge from their measured bounds
    The resolver captures mutable visibleAgentNotes and expandedKeys references at
    diffSectionGeometry.ts:137-165. Bounds are measured immediately, but rows are rebuilt
    later and paired by index at copySelection.ts:322-324. Mutating those inputs before first
    access can misalign copied rows. Snapshot the inputs when creating the resolver.
```

PR #530, opened 2026-07-12
Odin support

dviramontes:
## Context
- [Odin lang](https://odin-lang.org/)

## What changed

- Recognize `.odin` files during diff language detection
- Register Shiki's Odin TextMate grammar through Pierre's custom-language API
- Cover both extension detection and rendered syntax token colors
- Updates Changelog

## Why

Pierre's pinned Shiki 3 bundle does not include the Odin grammar, so `.odin` diffs previously fell back to plain text. Loading the current grammar separately preserves Pierre as the renderer while enabling syntax-aware Odin diff viewing.

## Validation
- Unit test 

- Manual file-pair review of representative Odin source

```shell
  printf 'package main\n\nmain :: proc() {}\n' > /tmp/before.odin
  printf 'package main\n\nmain :: proc() {\n  message := "hello"\n}\n' > /tmp/after.odin
  bun run src/main.tsx -- diff /tmp/before.odin /tmp/after.odin
  ```

Screenshot 2026-07-12 at 3 45 58 PM

### DIFF
--- a/.changeset/odin-highlighting.md
+++ b/.changeset/odin-highlighting.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Highlight Odin source files in diff views.
--- a/bun.lock
+++ b/bun.lock
@@ -6,6 +6,7 @@
       "name": "hunk",
       "dependencies": {
         "@pierre/diffs": "1.2.2",
+        "@shikijs/langs": "4.0.2",
         "bun": "^1.3.14",
         "chokidar": "^4.0.3",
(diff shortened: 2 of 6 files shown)


benvinegar:
Any chance of getting this accepted upstream in Shiki?

dviramontes:
> Any chance of getting this accepted upstream in Shiki?

Did not explore that, but I can give it a shot and report back.

benvinegar:
I'm also going to see if we can make this extensible via the new extension system.

PR #531, opened 2026-07-13
feat(watch): replace 250 ms watch polling with evented filesystem observation

elucid:
## Summary

Replace `--watch`'s 250 ms Git polling loop with filesystem event hints plus the existing authoritative Git signature. Watch remains continuous—passively viewed sessions stay fresh—while idle sessions stop spawning several Git commands four times per second.

PR #520 identified the same important unattended-session cost and proposed pausing polling after UI inactivity. This PR takes a different approach that preserves watch semantics and keeps filesystem observation independent of keyboard and mouse activity.

## Why a hybrid watcher

- **macOS and Windows:** native recursive `fs.watch`, where the OS can observe a whole tree with a small constant number of registrations.
- **Linux and portable fallback:** Chokidar recursion pruned with ignored roots reported by Git, avoiding a native inotify registration for every directory—including ignored dependency trees.

This is a standard shape for development tooling: Turbopack similarly uses efficient recursive observation on macOS/Windows and selective non-recursive observation on Linux. Git remains Hunk's source of truth; events only tell Hunk when it may be worth recomputing the signature.

A 2-second readiness deadline and `ENOSPC`/`EMFILE` handling close partial observers and fall back to a safe 2-second poll. A healthy observer retains a 10-second safety check for missed events or unusual filesystems.

## macOS and Linux benchmark summary

The current [benchmark report](https://github.com/modem-dev/hunk/blob/elucid/file-watch/docs/watch-benchmark-final.md) compares exact compiled base and candidate SHAs with Bun 1.3.14 across a **little repo (1,260 subdirectories)** and **big repo (25,223 subdirectories)**.

| Platform | Startup change | 120-second idle CPU | 120-second Git calls |
| --- | ---: | ---: | ---: |
| macOS arm64 | +27 to +61 ms | **4.4–6.4× less** | **35–36× fewer** |
| Linux x64 | +21 to +24 ms | **1.8–5.4× less** | **36× fewer** |

All measured refresh trials were correct for tracked writes, atomic saves, and relevant untracked creation. Candidate p95 remained below 610 ms. The report contains the frozen provenance, raw-artifact checksums, full startup/idle/latency tables, and disclosed Linux recovery evidence.

## Windows status

Windows unit and focused watch tests pass, and native/Chokidar observer probes both reach ready. We were not able to complete the full TUI benchmark matrix yet:

- Windows ARM64 is blocked by Bun 1.3.14/OpenTUI FFI initialization.
- Windows x64 GitHub runners are blocked by a surviving ConPTY child during headless teardown.

Windows performance is therefore not inferred from macOS, and no Windows startup, CPU, memory, or latency claim is made. Follow-up work will fix the Windows TUI harness and run the same two-fixture campaign for both the native production path and forced Chokidar comparison.

## How it works

```text
filesystem event (hint)
  -> debounce (200 ms quiet / 1 s maximum)
  -> recompute authoritative signature
  -> refresh only when it changed
```

- Backend-neutral watch plans describe worktree, metadata, direct-file, patch, and sidecar targets.
- Git plans handle ignored roots, linked-worktree metadata, refs, the index, and stash state.
- A deterministic controller serializes checks and refreshes and guarantees one trailing check for events received during a refresh.
- The old 250 ms `setInterval` path is removed from the UI.

## Benchmark harness

The WIP harness lives on [`elucid/watch-benchmark-harness`](https://github.com/modem-dev/hunk/tree/elucid/watch-benchmark-harness). It freezes source/binary provenance, builds portable directory-heavy fixtures, runs 120-second idle cohorts and TUI latency trials, and retains raw JSON/log artifacts. The campaign is intentionally too heavy for normal CI; once the Windows blockers are resolved, the branch will be adjusted for manual/ad-hoc runs rather than every PR push.

## Scope and verification

- Evented coverage: Git working-tree/staged/range/show/stash reviews, direct diffs, patch files, difftool inputs, and agent-context sidecars.
- Jujutsu and Sapling currently use the 2-second polling fallback pending provider-specific plans.
- Focused watch/controller/observer tests, source and compiled PTY regressions, typecheck, lint, packaging, and Nix dependency packaging were exercised during implementation.
- Adds `chokidar@^4.0.3` as a bundled runtime dependency and includes patch changesets.


benvinegar:
Nice

elucid:
Frozen watch benchmark campaign: `watch-20260714T185115Z-b04ed2b0b-h4f35a0ef`

- base: `04ed2b0bd51aae633ab94b3a6d157d5d5e568dd0`
- candidate: `4f35a0ef67f9d8094b4b3ff5bd74c3b74940dd57`
- harness: `bcfb232b4e83e3c8a07834c5e114e7f0f364e8d5`
- modem fixture: `85caaceeff750eeceb2e26a7336c8aef1d8372f5`

The candidate merge-base equals the frozen base. Any revision change invalidates this campaign ID.

benvinegar:
I think a follow up would be to document the watch* files more (like at the top of the file). And maybe some guidance for how other VCSs can tap in.

PR #536, opened 2026-07-14
revert #485 and bump diffs to 1.3.0-beta.7

scaryrawr:
Revert #485 which added custom extensions for mts/cts. diffs now natively supports these extensions as of 1.3.0-beta.7


benvinegar:
@scaryrawr Can we wait until this leaves beta for diffs?

PR #540, opened 2026-07-14
feat: auto-discover target-keyed agent-context sidecars

alleneubank:
## Summary

- Auto-discover **target-keyed** agent-context sidecars: `.hunk/agent-context.<targetId>.json` for the current review target (working tree, staged, range, show, stash-show + sorted pathspecs)
- **Never** auto-load bare `.hunk/agent-context.json` (that was the confusing partial-attach case). Explicit `--agent-context` / config `agent_context` still load any path, including a legacy bare name
- Keep explicit/configured paths strict; conventional keyed path is best-effort (missing/malformed → no notes, review continues)
- Show agent notes when a sidecar actually loads; `--no-agent-context` opt-out; exclude `.hunk/` metadata from untracked review noise
- Document the convention (README, agent-workflows, hunk-review skill) + minor Changeset

## Why (response to review)

@benvinegar called out that unscoped auto-discovery of a fixed bare sidecar can load notes against a different review and partially apply them. This revision keys discovery on **review target identity in the filename** (not patch content hash, which would thrash on every working-tree edit). Same pure function for discovery, watch, and agent skill guidance.

`targetId` = first 12 hex of SHA-256 over a canonical target string. Users never type it; agents write the conventional path for the same CLI args they open the review with.

## User impact

| Before | After |
|--------|--------|
| Bare `.hunk/agent-context.json` auto-loaded on every `hunk diff` / show / range | Only `.hunk/agent-context.<id>.json` for **this** target auto-loads |
| One shared file for all reviews | Multiple sidecars can coexist per range/show/WT |
| Partial attach across mismatched targets | Wrong-id file ignored; no dual bare+keyed auto-read |

## Verification

- [x] `bun run typecheck`
- [x] focused: `bun test src/core/agentContextPath.test.ts src/core/config.test.ts src/core/agent.test.ts` (78 pass)
- [x] `bun run typecheck` on this branch tip

## Notes for cherry-pick

Single feat commit on current `main`. No SPEC/BRIEF/LOOP campaign docs on this PR.

### DIFF
--- a/.changeset/agent-context-auto-discovery.md
+++ b/.changeset/agent-context-auto-discovery.md
@@ -1,0 +1,8 @@
+---
+"hunkdiff": minor
+---
+
+Auto-discover `.hunk/agent-context.json` so agent review notes appear in `hunk diff`
+with no flags. Adds an `agent_context` config key (path resolved against the repo root)
+and a `--no-agent-context` opt-out, shows agent notes by default when a sidecar loads,
+and keeps hunk's own `.hunk/` metadata out of untracked working-tree review noise.
--- a/.gitignore
+++ b/.gitignore
@@ -42,8 +42,20 @@ report.[0-9]_.[0-9]_.[0-9]_.[0-9]_.json
 tmp
 .hunk/latest.json
 .hunk/config.toml
+.hunk/agent-context.json
+.hunk/review-state.json
+.hunk/review-comments.json
 .pi/
 autoresearch.jsonl
 autoresearch.ideas.md
 result
+
+# vscode extension package build + downloaded test editor
+editors/vscode/out
+editors/vscode/.vscode-test
+# `vsce package` output — built on demand, never committed.
+editors/vscode/*.vsix
+# The end-to-end suite turns the fixture workspace into a real repo for the run.
+editors/vscode/fixtures/workspace/.git
+editors/vscode/fixtures/workspace/.hunk
 .video-work/
--- a/README.md
+++ b/README.md
@@ -9,10 +9,16 @@ Hunk is a review-first terminal diff viewer for agent-authored changesets, built
 [![MIT License](https://img.shields.io/badge/License-MIT-blue.svg?style=for-the-badge)](LICENSE)
 
 - multi-file review stream with sidebar navigation
+- per-file viewed progress with keyboard navigation and repo-local persistence
 - inline AI and agent annotations beside the code
 - split, stack, and responsive auto layouts
 - watch mode for auto-reloading file and Git-backed reviews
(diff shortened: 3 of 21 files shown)


benvinegar:
I think this is a good idea, but I'd want it to somehow match the changeset/rev range somehow. e.g. the next time you run `hunk diff`, it could be a different diff, and the comments would be stale / not apply / could partially apply (confusing).

So I think we'd want some kind of id in the sidecar json filename.

alleneubank:
@benvinegar — thank you for the review on the earlier auto-discovery shape.

We took your suggestion: discovery is now **target-scoped via the filename**, not a single bare `.hunk/agent-context.json`.

**Convention:** `.hunk/agent-context.<targetId>.json`  
where `targetId` is the first 12 hex chars of SHA-256 over a canonical review-target string (working-tree / staged / range+expression / show+ref / stash-show+ref + sorted pathspecs). Same pure function for load, watch, and agent guidance.

**Guarantees:**
- A sidecar written for `hunk diff main...HEAD` does **not** auto-load on bare `hunk diff`
- Two ranges in one repo can each keep a sidecar without clobbering
- Explicit `--agent-context` / config still load any path (including a legacy bare name if someone wants that)
- Missing keyed file → review works, no notes (best-effort)

We deliberately did **not** put a full patch content hash in the filename (working-tree thrash / agents rewriting the name every edit). Content drift still soft-fails via existing range matching.

Happy to adjust further if you want a different canonical payload or id length.

benvinegar:
Ok, apologies, I'll need some time to think on this - I'm still following

Novohudonossor:
+1 from an external tool author, with one concrete ask.

We build ReviewGate — AI code review with team rules and a judge step, shipped as a merge-request bot and a local CLI. Today we export findings to hunk through the explicit --agent-context path and it works great: a 7-finding review renders as notes under their lines, author = generator → judge.

We'd like to write the keyed sidecar directly, so that hunk diff --staged after a review just picks it up — no flag, no stale file. The convention in this PR — target identity in the filename, never auto-load the bare file — is exactly what a writer wants: I'd rather miss than attach notes to the wrong review, which is the concern you raised above, @benvinegar.

The one gap for a non-agent writer: a plain binary doesn't read the hunk-review skill, so it needs a stable way to get the conventional path for a given review target — either a helper (hunk agent-context path --staged / hunk agent-context path diff origin/main printing .hunk/agent-context.<targetId>.json) or the canonical target string + hash documented so tools can reproduce it byte-for-byte. With that, any reviewer/linter can drop notes into hunk without coordination.

Happy to be a test case for the convention once it lands.

PR #548, opened 2026-07-15
fix(ui): copy selection misaligns on wide (CJK) characters

endotakuya:
## Summary

Mouse-drag copy selection drifted whenever the selection crossed full-width characters (CJK, emoji). Mouse coordinates are terminal **cell** columns, but the copy path sliced rendered lines with `String.prototype.slice`, which counts UTF-16 **code units**. ASCII stays aligned (1 cell = 1 code unit), but each full-width character is 1 code unit and 2 cells, so every one before a selection endpoint shifted the copied text by one position.

- add `cellRangeToCharRange` in `src/ui/lib/text.ts` to convert an inclusive terminal-cell range into code-unit slice bounds, walking grapheme clusters so wide characters and emoji are never split
- convert cell columns to string indices before every slice in `renderCopySelectionText` (code-only and decorated paths, drag start and end) and in `expandSelectionPoint` (double/triple-click word and line selection)
- fix the file-header copy path too: `renderFileHeaderCopyText` now measures the gap and clamps by cells, matching `DiffFileHeaderRow`'s cell-based layout for wide-character filenames
- keep zero-width characters at a selection boundary so invisible characters round-trip through the clipboard
- add a patch changeset

### Repro

Drag-select the prefix `export const message = 'こんにちは` on this line:

```
export const message = 'こんにちは'; // greeting
```

That prefix is **29 UTF-16 code units but 34 terminal cells** (five full-width characters, +1 cell each), so slicing by code units overshoots by five cells:

| | Copied text |
|---|---|
| Before | `export const message = 'こんにちは'; //` |
| After | `export const message = 'こんにちは'` |

## Testing

- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- `bun run test:theme-contrast`
- `bun test`
- `bun run test:integration`
- `bun run test:tty-smoke`

### DIFF
--- a/.changeset/wide-char-copy-selection.md
+++ b/.changeset/wide-char-copy-selection.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Fix mouse-selection copy misalignment on lines with wide (CJK, emoji) characters: drag, double-click, and triple-click selections now convert terminal cell columns into string indices before slicing, so the copied text matches the selected cells exactly. File-header rows with wide-character filenames now copy with the same cell alignment, and invisible zero-width characters at a selection boundary round-trip through the clipboard.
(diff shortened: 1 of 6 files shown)


benvinegar:
Thanks for the thoughtful fix and thorough coverage! 🙏

*This comment was generated by Pi using OpenAI GPT-5.2*

PR #567, opened 2026-07-18
fix(ui): wrap agent note text by terminal cells

### LINKED ISSUE #566: Agent note text with CJK characters is truncated instead of wrapped (wrapText measures UTF-16 code units, not terminal cells)
## What happened?

Plain-text agent notes (inline note cards, hover popovers, and the static
pager's inline-note lines) truncate CJK text instead of wrapping it.

With a Japanese summary the inline card renders a single clipped line:

```
╭─ Claude note - src/ui/lib/agentPopover.ts R33 ─────────────────────────────╮
│ この注釈はわざと長い日本語で書いています。0.17.1では折り返されず途中で切.  │
╰────────────────────────────────────────────────────────────────────────────╯
```

Depending on the box width, the same note either loses its tail (a short
summary passes the wrap check as "fits on one line" and gets clamped at
render time) or silently loses text from the middle (a long summary is
hard-split by code units into lines up to 2× the box width, and the
renderer clips every line — the clipped lines still read as continuous
prose, so the loss is easy to miss). Emoji can also be split through a
surrogate pair, so a line can end in a lone surrogate (renders as U+FFFD).

An English note of the same length wraps correctly — ASCII is one cell per
code unit, so the mismatch never shows.

## Steps to reproduce

Smallest repro, directly against `wrapText` (repo checkout):

```ts
// bun run repro.ts
import { wrapText } from "./src/ui/lib/agentPopover";

// 62 chars, 124 terminal cells; Japanese has no spaces to break on
const ja = "このサマリーはリトライループの遅延が試行回数の増加とともに無制限に伸びる問題を説明します。全角文字だけで構成された文章です。";
console.log(wrapText(ja, 40));
// -> 2 lines of 40 *code units* each = 80 cells each (2× the width)

console.log(wrapText("🎉".repeat(30), 25));
// -> line 1 ends with a lone high surrogate
```

In the TUI: open any diff, add a note with a long Japanese summary
(`hunk session comment add … --summary "…"`), and compare with an English
summary of similar length. The Japanese card is clipped with a trailing
`.`; the English card wraps to multiple lines with nothing lost.

## Expected behavior

Wrapped lines never exceed the requested width in terminal cells, no text
is silently dropped (including degenerate widths narrower than one
cluster), and grapheme clusters (emoji, surrogate pairs) are never split.

## Suspected cause

`wrapText` in `src/ui/lib/agentPopover.ts` measures with `String#length`
and splits with `String#slice` (UTF-16 code units), while the note
renderers clamp each line with the cell-aware `fitText` / `padText` from
`src/ui/lib/text.ts`. Where code units ≠ cells the two layers disagree,
and the clamp turns the mismatch into truncation:

- `word.length > width` / `next.length <= width` — a Japanese summary is
  one long "word" (no spaces), so a 66-char / 132-cell summary passes
  checks that are really about cells
- `word.slice(offset, offset + width)` — hard-splits by code units,
  producing over-wide lines and cutting surrogate pairs in half

Diff-body rendering and STML notes already wrap by cells via
`measureTextWidth` / `sliceTextByWidth` — the same cell-width helpers in
`src/ui/lib/text.ts` that back #548's copy-selection fix — so the
plain-text note path is the remaining code-unit path. Same class of bug
as #324, which covered the diff body.

Two smaller code-unit spots remain in the note UI (the draft textarea
height estimate and the card's title border padding) — different symptoms,
so I've left them out of scope here; happy to file separately if useful.

## Version

hunk 0.17.1 (Homebrew) on macOS, Ghostty. Also reproduced against current
main (b580edd) with the snippet above.

---

A fix is ready (reusing `measureTextWidth` / `sliceTextByWidth` inside
`wrapText`, plus regression tests) — I'm opening a PR right after filing
this issue.

kataokatsuki:
## Summary

Plain-text agent notes (inline cards, hover popovers, static pager
inline-note lines) wrap their text with `wrapText`, which measures with
`String#length` and splits with `String#slice` — UTF-16 code units. The
renderers then clamp each line by terminal cells (`fitText` / `padText`),
so for CJK text (1 code unit = 2 cells) the wrapped lines come out up to
2× the box width and the clamp truncates them, silently dropping text.
Code-unit splitting can also cut an emoji's surrogate pair in half
(renders as U+FFFD).

- measure words with `measureTextWidth` and track the current line's
  cell width alongside the string, so line-fit checks compare cells to
  cells
- hard-split over-wide words with `sliceTextByWidth`, which walks
  grapheme clusters — wide characters and emoji are never split
- when the width is narrower than a single cluster, keep the remainder
  on one line for the renderer's `fitText` to clamp instead of dropping
  it, and decide "line is empty" by string length so zero-width
  characters can't produce over-wide lines
- add regression tests: CJK wrapping, mixed EN/JA text, odd widths,
  degenerate width 1, surrogate-pair and ZWJ emoji

Closes #566

### Repro

With a 40-cell box and a 62-char / 124-cell Japanese summary:

| | wrapText output |
|---|---|
| Before | 2 lines of 40 code units = 80 cells each; the renderer clips both lines, silently dropping ~half the text |
| After | 4 lines of ≤ 40 cells; nothing dropped |

English output is unchanged — wrapping of ASCII input is byte-identical
before and after.

## Testing

All run in a clean Linux container (oven/bun + git), matching CI's
ubuntu-latest environment:

- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- `bun test src` — 1,140 tests, 0 fail
- `bun run test:tty-smoke` — 9 pass, identical to a baseline run without
  this change (the harness needs util-linux `script`, which macOS lacks)

### DIFF
--- a/.changeset/wrap-agent-notes-by-cells.md
+++ b/.changeset/wrap-agent-notes-by-cells.md
@@ -1,0 +1,8 @@
+---
+"hunkdiff": patch
+---
+
+Wrap plain-text agent notes by terminal cells instead of UTF-16 code
+units, so CJK and emoji text wraps correctly instead of being truncated
+with silent content loss. Long unbroken words split on grapheme
+boundaries, so wide characters and surrogate pairs are never cut apart.
--- a/src/ui/lib/agentPopover.ts
+++ b/src/ui/lib/agentPopover.ts
@@ -1,11 +1,11 @@
 import { sanitizeTerminalLine } from "../../lib/terminalText";
-import { fitText } from "./text";
+import { fitText, measureTextWidth, sliceTextByWidth } from "./text";
 
 function clamp(value: number, min: number, max: number) {
   return Math.min(Math.max(value, min), max);
 }
 
-/** Wrap plain text to a fixed terminal width, breaking long tokens when needed. */
+/** Wrap plain text to a fixed terminal-cell width, breaking long tokens when needed. */
 export function wrapText(text: string, width: number) {
   if (width <= 0) {
     return [""];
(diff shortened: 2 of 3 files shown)


kataokatsuki:
Thanks for the quick merge!

Issue #556, opened 2026-07-17
Every invocation leaks a 3.6MB dylib into tmpdir


franky47:
### What happened?

Each run of `hunk` extracts its embedded `libopentui.dylib` into `$TMPDIR` as a hidden file named `.{hash}-00000000.dylib` (3.6MB). The file gets a unique random name on every run and is never deleted, even when hunk exits cleanly.

This adds up fast when hunk is used as a git pager. With lazygit spawning `git diff | hunk pager` on every refresh, my machine accumulated 81,653 of these files (about 190GB) in three weeks, at a rate of roughly 18 files per minute while lazygit was open (which I often leave open in a cmux pane, one per workspace).

The files identify as OpenTUI's native library:

```
$ codesign -dv .5ccbd3e7f7bf5fff-00000000.dylib
Identifier=libopentui.dylib
Format=Mach-O thin (arm64)
Signature=adhoc
```

This looks like Bun's single-file executable behaviour: embedded native addons are extracted to TMPDIR at startup, and the randomized filename prevents reuse across runs. Since cleanup doesn't happen either, every invocation adds a new copy.

Reproduced on v0.17.0 (Homebrew) and v0.17.1 (release binary `hunkdiff-darwin-arm64`), macOS 15 (Darwin 24.6.0), Apple Silicon.

### Steps to reproduce

```bash
#!/usr/bin/env bash
set -euo pipefail

REPRO_TMP="$(mktemp -d)"
trap 'rm -rf "$REPRO_TMP"' EXIT

DIFF=$(cat <<'EOF'
diff --git a/foo.txt b/foo.txt
index 0000000..1111111 100644
--- a/foo.txt
+++ b/foo.txt
@@ -1 +1 @@
-hello
+world
EOF
)

for i in 1 2 3 4 5; do
  TMPDIR="$REPRO_TMP/" hunk pager <<<"$DIFF" > /dev/null 2>&1
done

find "$REPRO_TMP" -name '.*.dylib' -exec ls -la {} +
```

Output on my machine: 5 runs leave 5 files, 18MB total.

```
-rw-r--r--@ 1 franky staff 3758912 .9bdbb2fbe532eefb-00000000.dylib
-rw-r--r--@ 1 franky staff 3758912 .9bdbb2fbeeafdefb-00000000.dylib
-rw-r--r--@ 1 franky staff 3758912 .9bdbb2fbf7afcfbb-00000000.dylib
...
```

### Expected behavior

[...]

PR #568, opened 2026-07-18
fix(cli): sweep stale Bun-extracted temp libraries at startup

WakaTaira:
## What

Bun single-file executables extract their embedded native libraries into the OS temp directory under a randomized hidden name (`.{16hex}-{8hex}.so|dylib|dll`) on every launch, and never reuse or remove them (oven-sh/bun#30962). For hunk this means every invocation leaks one copy of the OpenTUI native library — #556 documented 81,653 files (~190 GB) in three weeks via the lazygit pager integration; I hit a full 16 GB tmpfs overnight on Linux with an agent polling `hunk session` every 2 s.

This PR adds a best-effort startup sweep that reaps hunk's stale leaked artifacts until Bun's extraction dedupe (oven-sh/bun#29587) ships — at which point this module can be deleted outright.

## How

- New `src/core/tmpArtifactSweep.ts` (DI-friendly, colocated tests): removes hidden `.{16hex}-{8hex}.(so|dylib|dll)` regular files in the temp directory root, owned by the current user and older than one hour.
- Rate-limited to one directory scan per hour via a stamp file, claimed before the scan so concurrent starts don't all readdir.
- Hardened for shared /tmp: the stamp must be a regular file owned by the current user — a symlinked or foreign-owned stamp aborts the sweep; first creation is an exclusive write.
- Wired in `main()`: the sweep starts before argument parsing, and short-lived commands await it right before `process.exit` (which would otherwise drop the pending work). The app and daemon paths never block on it.
- All errors swallowed, no output. Opt out with `HUNK_DISABLE_TMP_SWEEP=1`.

## Verification

- `typecheck` / `test` / `lint` / `format:check` all pass (9 new unit tests).
- End-to-end on Linux: a compiled `hunk` binary leaks exactly one new artifact per `--help` run (reproducing the bug); with this patch it removes planted stale artifacts, keeps fresh ones, honors the opt-out, and respects the rate-limit stamp.

Refs #556.


WakaTaira:
Rebased onto the latest `main` to resolve the conflict with the recent module reorg (`app/startup`, `session/*`) — the sweep itself is unchanged. `typecheck`, `test` (1767 pass), `lint`, and `format:check` all pass locally.

One honest observation: at some point the leak seems to have stopped occurring on my machine — I no longer notice new artifacts accumulating. I haven't investigated why (I'm on 0.17.7, which predates #590's headless lazy-load, so that shouldn't be the reason), so it may be environmental. Sharing as anecdotal signal in case it's useful when deciding how this and #577 fit together.

benvinegar:
> One honest observation: at some point the leak seems to have stopped occurring on my machine — I no longer notice new artifacts accumulating.

I believe this is because I fixed a bug where it was firing up OpenTUI on every `hunk session` command, which was totally unnecessary.

I wanted to see how this went before adding the sweep PR or my other PR. Sounds like it's making a difference.

WakaTaira:
Caught the branch up with `main` — it had drifted ~109 commits behind, across the 0.18/0.19 releases. I went with a merge rather than a rebase this time to keep the existing discussion and comment anchors intact; happy to rebase to a linear history instead if you prefer.

The only conflict was in `src/main.tsx`, where the sweep's `exitAfterSweep(0)` overlapped with the new `preserveAnsiStyle` passthrough option — resolved by keeping both. The sweep module itself is unchanged. One small follow-up commit: the newer `extension manage` and `self-update` commands exited via bare `process.exit`, so I routed them through `exitAfterSweep` like the other short-lived startup kinds.

`typecheck`, `test` (1555 pass, 0 fail), `lint`, `format:check`, and `deps:check` all pass locally.

And since I'd gone quiet on syncing for a while: this PR is still very much on my radar. No rush on your end — watching how the lazy-load fix plays out before deciding how this and #577 land together makes sense to me. Whenever review comes, I'll turn changes around quickly.

WakaTaira:
One more note: I think I'd half-misread the wait-and-see here as meaning I should hold off on my side too. This sync turned up upstream changes that actually needed folding into the sweep's exit path (the new short-lived commands), so from here I'll keep the branch checked and caught up with `main` rather than waiting.

PR #581, opened 2026-07-20
fix(review): save draft notes exactly once under rapid Ctrl+S

### LINKED ISSUE #580: Rapid save (Ctrl+S or double-click) duplicates a draft note
## Summary

Pressing `Ctrl+S` (or double-clicking the `Save (^S)` button) more than once in quick succession while saving a draft review note saves the same note multiple times. Duplicates saved within the same millisecond also share one identical `user:${Date.now()}` id.

## Steps to reproduce

1. Open any diff, e.g. `hunk diff before.ts after.ts`.
2. Press `c` to start a draft note and type any text.
3. Press `Ctrl+S` twice quickly (fast enough that both key events arrive in one stdin chunk).

Note: this is a race — a manual double-press only triggers it when both key events land in one stdin chunk (holding the key down so it auto-repeats makes it easy). Sending `\x13\x13` in one write over a PTY reproduces it deterministically.

Observed: two identical notes are rendered (`Your note 1/2` / `Your note 2/2`), both with the same id (e.g. `user:1784558001062`).

Expected: exactly one note is saved; extra presses are no-ops.

## Root cause

`saveDraftNote` (`src/ui/hooks/useReviewController.ts`) guards re-entry only through the `draftNote` state closure. React state is a per-render snapshot, so `setDraftNote(null)` takes effect only after the next render; key events delivered in one stdin chunk run synchronously against the same still-non-null draft, and each call appends a note. All three save entry points (textarea keydown, app-level `Ctrl+S` fallback, `Save (^S)` button `onMouseUp`) converge on this function.

The shared id then breaks id-addressed consumers: `removeUserNote` / `removeLiveComment` (and the note's `[x]` close button) delete both duplicates at once, `hunk session comment list` exposes the duplicated `noteId` to agents, and React keys derived from `annotation.id` collide.

## Suggested fix

1. Add a render-cycle-independent re-entry guard to `saveDraftNote` (a ref tracking the already-saved draft id), the standard pattern for this bug class (cf. [TanStack Query discussion on duplicate mutations](https://github.com/TanStack/query/discussions/10041), [React docs: State as a Snapshot](https://react.dev/learn/state-as-a-snapshot)).
2. Make saved note ids collision-free (e.g. a monotonic counter suffix) so `user:*` ids stay unique within one millisecond.

## Environment

- Version: v0.17.3 (latest release; also reproduces from source at main `53fcb2c`)
- macOS (Darwin 25.5.0), reproduced in a real PTY against the released binary, with both Ctrl+S key events delivered in one stdin chunk

endotakuya:
## Summary

- Guard `saveDraftNote` re-entry with a ref keyed by the draft id, so save events that arrive before the draft-clearing state update commits are no-ops.
- Suffix saved note ids with a monotonic counter so `user:*` ids stay unique within one millisecond.

Closes #580.

## Why

React state is a per-render snapshot, so the old `if (!draftNote)` guard could not reject a second Ctrl+S delivered in the same stdin chunk (or a `Save (^S)` double-click): both events saw the still-non-null draft and each appended a note. The duplicates also shared one `user:${Date.now()}` id, so id-addressed removal (the `[x]` close button, `hunk session comment remove`) silently deleted both copies.

## Testing

- `bun run typecheck`
- `bun run lint`
- `bun test src/ui/hooks/useReviewController.test.tsx` (24 passed)
- `bun test ./test/pty/notes.test.ts` (13 passed, includes the new rapid Ctrl+S regression test, which also reproduces the bug against the released v0.17.3 binary via `HUNK_TEST_EXECUTABLE`)
- `bun test ./src ./packages ./scripts ./test/cli` (1,180 passed; `./test/session` fails on this machine identically on `origin/main` due to a live local hunk daemon, so relying on CI there)

### DIFF
--- a/.changeset/rapid-save-duplicate-note.md
+++ b/.changeset/rapid-save-duplicate-note.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Rapidly pressing Ctrl+S (or double-clicking Save) while saving a draft note no longer saves the same note twice, and saved note ids stay unique even within one millisecond.
--- a/src/ui/hooks/useReviewController.test.tsx
+++ b/src/ui/hooks/useReviewController.test.tsx
@@ -658,6 +658,59 @@ describe("useReviewController", () => {
     }
   });
 
+  test("rapid duplicate saves persist exactly one user note with a unique id", async () => {
+    const { controllerRef, setup } = await renderReviewController([createAlphaFile()]);
+
+    try {
+      await flush(setup);
+
+      await act(async () => {
+        expectValue(controllerRef.current).startUserNote();
+        expectValue(controllerRef.current).updateDraftNote("Save me once.");
+      });
+      await flush(setup);
+
+      // Coalesced Ctrl+S key events invoke save twice before the draft-clearing
+      // state update commits; only the first call may persist a note.
+      let firstSavedId: string | null = null;
+      let secondSavedId: string | null = null;
+      await act(async () => {
+        const controller = expectValue(controllerRef.current);
+        firstSavedId = controller.saveDraftNote()?.id ?? null;
+        secondSavedId = controller.saveDraftNote()?.id ?? null;
+      });
+      await flush(setup);
+
+      expect(firstSavedId).toStartWith("user:");
+      expect(secondSavedId).toBeNull();
+      expect(expectValue(controllerRef.current).userNotesByFileId.alpha).toHaveLength(1);
+
+      // A follow-up draft saved within the same millisecond still gets a unique id.
+      await act(async () => {
+        expectValue(controllerRef.current).startUserNote();
+      });
+      await flush(setup);
+      await act(async () => {
+        expectValue(controllerRef.current).updateDraftNote("Save me too.");
+      });
+      await flush(setup);
+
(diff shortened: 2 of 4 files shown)


benvinegar (changes requested):
The fix itself looks good, and I reproduced the issue and verified the fix in tmux. Could you tighten up these two tests before merging?

- In the PTY test, `session.type("\x13\x13")` sends the two characters separately with a short delay, so it doesn't guarantee the coalesced input described in the comment. `await session.write("\x13\x13")` would send both bytes in one PTY write and make the regression test deterministic.
- In the controller test, please freeze or mock `Date.now()` for both saves. As written, the IDs may differ because the clock advanced between renders, so it doesn't actually prove IDs stay unique within the same millisecond.

*This comment was generated by Pi using OpenAI GPT-5.4*

benvinegar (approved):
I pushed the two test fixes: the PTY test now sends both Ctrl+S bytes in one write, and the controller test holds `Date.now()` constant while checking the exact sequence IDs. The focused tests, typecheck, lint, and formatting all pass.

*This comment was generated by Pi using OpenAI GPT-5.4*

endotakuya:
@benvinegar Thank you for the review, the test fixes, and the merge!

Issue #579, opened 2026-07-20
Scroll wheel use freezes screen


peterlandry:
### What happened?

I suspect scroll events from a scroll wheel aren't being debounced properly. If I'm viewing diffs (with or without `--watch`), doing a large quick scroll freezes the UI. There are some partial torn redraws that happen, then eventually it settles.

### Steps to reproduce

Open a diff that requires scrolling. Do a large scroll by "flicking" the scroll wheel.

### Expected behavior

Scrolling to happen, and stop at the top or bottom.

### Version

0.17.1

benvinegar:
I took a look and came up empty; I can't reproduce. Can you give me more information? e.g.

1. OS/architecture, terminal/version, tmux/SSH, etc
2. Wrap/layout/theme configuration and whether agent notes are present
3. Approximate file count and changed-line count
4. A shareable patch or synthetic reproduction
5. A short screen recording.

peterlandry:
1. MacOS / ghostty build 16124 / herdr
2. No agent notes, default theme
3. Seems to happen with any counts. I think last instance was 3 or 4 changed files, a few hundred lines of code change
4. TODO
5. TODO

zivattias:
I can also confirm this behavior, using Ghostty + Herdr.

EDIT: Only when using `hunk diff --watch`

benvinegar:
So, I've seen my colleague @tridha643 experience this with herdr (I was sitting over his shouler). So that's 3 people experiencing this with `herdr`.

His agents were maxing out the CPU on his machine (100% utilization). What does your system utilization look like when this happens?

benvinegar:
And what version of hunk btw? Are you using 0.17.3?

kazu728:
https://github.com/user-attachments/assets/d8a403a8-474c-4251-ad8f-90b046cefc1a

```sh
❯ hunk --version
0.17.3

❯ bun --version
1.3.13
```
Same issue here. I'm not sure which language triggered the original report, but for Japanese text specifically, I suspect the bottleneck might not be Herdr's own overhead — it could be the regex, which is fairly expensive and may be getting invoked on every scroll event, possibly causing the lag.

benvinegar:
We think the fix is in this release: https://github.com/modem-dev/hunk/releases/tag/v0.17.4

Give it a shot?

> which is fairly expensive and may be getting invoked on every scroll event, possibly causing the lag.

Yep, probably

benvinegar:
[...]

PR #586, opened 2026-07-23
perf(ui): optimize terminal cell width measurement (#579)

kazu728:
In measureSanitizedTextWidth (src/ui/lib/text.ts), non-ASCII text falls to string-width, which classifies every grapheme cluster via `/^\p{RGI_Emoji}$/v`. So plain Japanese pays emoji-detection cost too, and that v-flag regex is pathologically slow on Bun.

https://github.com/user-attachments/assets/60005822-e00c-43ca-8854-1fb398db6e0e

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "chokidar": "^4.0.3",
     "commander": "^14.0.3",
     "diff": "^8.0.3",
+    "get-east-asian-width": "^1.5.0",
     "shell-quote": "1.8.4",
     "string-width": "^8.2.1",
     "zod": "^4.3.6"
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -99,6 +99,7 @@
     "chokidar": "^4.0.3",
     "commander": "^14.0.3",
     "diff": "^8.0.3",
+    "get-east-asian-width": "^1.5.0",
```
kazu728, line 103: Relying on implicit transitive dependencies can cause confusion, so I added get-east-asian-width as an explicit dependency. It was already included transitively via string-width.

PR #597, opened 2026-07-25
fix: remove bundled bun runtime dependency

### LINKED ISSUE #518: `hunkdiff` depends on npm `bun`, causing blocked Bun lifecycle script warnings during global install
## Summary

When installing or updating `hunkdiff` globally using Bun on Windows, Bun reports a blocked lifecycle script for the `bun` package:

```text
.\node_modules\bun @1.3.14
 » [postinstall]: node install.js
```

This appears to be because `hunkdiff` declares `bun` as a runtime dependency. In my global Bun install, `bun why bun` shows `hunkdiff` as the package introducing this dependency.

The confusion is that I already have Bun installed as the active runtime, but installing `hunkdiff` globally creates a separate nested `node_modules\bun` dependency whose `postinstall` script is blocked by Bun’s lifecycle-script security model.

## Environment

* OS: Windows 11 Pro
* Shell: PowerShell
* Bun version observed: `1.3.14`
* Bun executable path:

```powershell
where.exe bun
# C:\Users\user\.bun\bin\bun.exe
```

* Bun global package project:

```text
C:\Users\user\.bun\install\global
```

## Observation

Running:

```powershell
bun pm -g untrusted
```

shows:

```text
bun pm untrusted v1.3.14

.\node_modules\bun @1.3.14
 » [postinstall]: node install.js

These dependencies had their lifecycle scripts blocked during install.

If you trust them and wish to run their scripts, use `bun pm trust`.
```

Checking why `bun` is present:

```powershell
Push-Location "$env:USERPROFILE\.bun\install\global"
bun why bun
```

shows:

```text
bun@1.3.14
  └─ hunkdiff@0.17.0 (requires ^1.3.14)
```

I previously observed the same pattern with an earlier `hunkdiff` version as well, where `hunkdiff@0.16.0` required `bun@^1.3.10`.

## Why this is confusing/problematic

1. `hunkdiff` is being installed with Bun, and Bun is already available on PATH.
2. `hunkdiff` still installs the npm `bun` package as a dependency.
3. Bun then blocks the nested `bun` package’s `postinstall` script.
4. The suggested remediation is to run something like:

```powershell
bun pm trust bun
```

But from a user perspective, it is unclear whether it is safe or necessary to trust and run the nested `bun` package’s lifecycle script just to use `hunkdiff`.

## Reproduction steps

```powershell
# Install hunkdiff globally using Bun
bun add --global hunkdiff

# Check blocked lifecycle scripts
bun pm -g untrusted

# Inspect why the bun package exists
Push-Location "$env:USERPROFILE\.bun\install\global"
bun why bun
```

## Expected behavior

Installing `hunkdiff` globally with Bun should ideally not leave the global Bun package project with a blocked lifecycle script warning for a nested `bun` dependency.

Ideally one of the following would be true:

* `hunkdiff` does not need to depend on the npm `bun` package at runtime when installed globally, or
* the `bun` dependency is moved out of runtime dependencies if it is only needed for build/dev/package-manager tasks, or
* the package documents why `bun` is required as a runtime dependency and whether users should run `bun pm trust bun`, or
* the install path avoids requiring users to make a trust decision for a nested Bun runtime package.

## Actual behavior

`hunkdiff` pulls in the npm `bun` package as a dependency, and Bun reports that the nested `bun` package’s `postinstall` script was blocked.

## Workaround

Removing `hunkdiff` removes the dependency chain pulling in `bun`:

```powershell
Push-Location "$env:USERPROFILE\.bun\install\global"
bun remove --global hunkdiff
bun pm -g untrusted
```

Alternatively, users can run:

```powershell
bun pm trust bun
```

but that explicitly trusts and runs the nested `bun` package’s lifecycle script, which may not be desirable without clarification from the package maintainers.

## Question

Is the npm `bun` package required as a runtime dependency for `hunkdiff`, or could it be moved to a dev/build-time dependency or otherwise avoided for global installs?

If it is required at runtime, could the documentation explain why `hunkdiff` depends on the npm-distributed Bun runtime package and whether users installing with Bun should trust its lifecycle script?

### Steps to reproduce

See above

### Expected behavior

See above

### Version

0.17.0

cyphercodes:
## Summary

- remove the npm `bun` package from `hunkdiff` runtime dependencies
- keep `@types/bun` as a dev dependency for Bun type coverage
- update Bun and Nix dependency locks plus a patch changeset

Fixes #518.

## Validation

- `bun install --ignore-scripts`
- `bun run format:check`
- `bun run typecheck`
- `bun test scripts/prebuilt-package-helpers.test.ts test/cli/entrypoint.test.ts`
- `bun run build:npm`
- `bun run check:pack`
- `git diff --check`
- `bun why bun` (reports no matching package)
- `bun pm untrusted` (no `node_modules/bun` lifecycle warning)

### DIFF
--- a/.changeset/remove-bun-runtime-dependency.md
+++ b/.changeset/remove-bun-runtime-dependency.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Remove the npm `bun` package from runtime dependencies so Bun installs do not report a blocked nested Bun lifecycle script.
--- a/bun.lock
+++ b/bun.lock
@@ -6,7 +6,6 @@
       "name": "hunk",
       "dependencies": {
         "@pierre/diffs": "1.2.2",
-        "bun": "^1.3.14",
         "chokidar": "^4.0.3",
         "commander": "^14.0.3",
         "diff": "^8.0.3",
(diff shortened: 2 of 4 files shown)


benvinegar:
y tho?

PR #598, opened 2026-07-26
fix(update): show Nix-aware upgrade guidance

### LINKED ISSUE #549: Don't notify about updates if installed through nix

Remove the banner indicating that I can update with `npm -i -g hunk` if I install hunk through nixpkgs

### Why?

Since I don't have hunk installed via npm, running that command doesn't do anything for me.

### How? (optional)

_No response_

benvinegar:
## Summary

- recognize Nix installs from either the package marker or a `/nix/store` executable
- show neutral Nix-specific update guidance instead of an npm command
- mark Hunk's flake package explicitly and keep Nix installs on stable release notices

Closes #549

## Testing

- `bun test src/core/updateNotice.test.ts`
- `bun test "$PWD/src/core"`
- `bun run typecheck`
- `bun run format:check`
- `bun run lint`
- tmux 3.6b smoke test with `HUNK_INSTALL_SOURCE=nix`; confirmed the live footer shows the Nix guidance and no npm command

*This PR description was generated by Pi using OpenAI GPT-5.2*

### DIFF
--- a/.changeset/wise-pandas-update.md
+++ b/.changeset/wise-pandas-update.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Show Nix-aware update guidance instead of suggesting an npm install command for Hunk's Nix package.
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -2,6 +2,7 @@
   bun,
   bun2nix,
   lib,
+  makeWrapper,
   ...
 }: let
   packageJson = lib.importJSON ../package.json;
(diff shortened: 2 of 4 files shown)


tristan957:
I would have expected this to be a build time flag that way the nix derivation and the Homebrew build set the flag. Alternatively they could patch the default value of the config option. The config option could also default to false, and on npm publish the pipeline could change the default value to false. 

If AI was used to come up with this PR, I think it may have taken the issue a little too literally.

PR #607, opened 2026-07-27
perf(ui): keep wrapped large diffs responsive

### LINKED ISSUE #600: Wrapped large diff remains blank on v0.17.6 while v0.13.2 renders
## Summary

Hunk v0.17.6 remains blank indefinitely when rendering a large diff with line wrapping enabled. The same diff renders and remains navigable on v0.13.2, although memory usage is high. Disabling wrapping on v0.17.6 restores sub-second rendering.

This appears distinct from the general macOS startup regression reported in #534: that original small-fixture reproduction no longer regresses on v0.17.6.

## Environment

- macOS 26.5.2, arm64
- Kitty 0.45.0
- Fish 4.8.0
- Native arm64 Hunk binaries built from the corresponding tagged Nix flakes
- Hunk config:

```toml
agent_notes = true
line_numbers = true
theme = "ember"
watch = false
wrap_lines = true
```

## Reproduction

From an already-running Fish shell in Kitty, in a branch with 141 changed files, 20,852 insertions, and 1,475 deletions:

```fish
hunk diff main
```

Git produces the underlying diff in approximately 0.02 seconds.

## Results

| Version / configuration | Result |
| --- | --- |
| v0.17.6, wrapping enabled | Blank after several minutes; sampled at 2m42s with approximately 1 GB physical footprint and the main thread idle in the event loop |
| v0.17.6, `--no-wrap` | Complete diff rendered in 0.63s; navigation responsive; approximately 345 MB RSS during the observed run |
| v0.17.6, wrapping enabled, scoped to one 648-line file | Rendered in 0.65s |
| v0.13.2, wrapping enabled | Complete diff rendered in 2.83–2.85s; navigation responsive; approximately 2.28 GB RSS |

The comparisons used the same machine, Kitty instance, repository, terminal dimensions, diff, and configuration. Each run used a private `HUNK_MCP_PORT`. The literal wrapper path `hd main` only forwards to `hunk diff main` and reproduced the same results.

## Expected behavior

Wrapped reviews should either:

1. render with bounded memory and responsive navigation, or
2. degrade gracefully with progressive or single-file rendering, or a clear size warning, instead of leaving a blank TUI indefinitely.

## Relevant existing work

- #405 covered high memory usage on large reviews and was closed after the large-review memory and performance work.
- #413 covered large-diff interaction latency and was closed after v0.15.3.
- #560 confirms `wrap_lines = true` is a supported preference.
- PR #427 notes that wrapped mode does not use file-section windowing because wrapped rows have variable heights, leaving wrapped reviews on the full render path.

This may be a regression or an uncovered scalability boundary in that wrapped full-render path.

mplibunao:
Fixes #600

## Summary

- Enable exact file-level windowing for wrapped reviews instead of mounting every file section.
- Highlight selected and visible files immediately, while deferring speculative neighbor/offscreen highlighting until navigation is idle.
- Share wrapped-row width budgets between rendering and geometry, including note-guide and add-note columns.
- Give no-hunk messages an explicit two-row body so window spacers and mounted content stay aligned.
- Add component, geometry, render-window, and PTY regressions for wrapped navigation and rapid scrolling.

## Why

Wrapping previously bypassed file-level windowing. On a large review this mounted the entire file stream and eagerly queued syntax highlighting on the main thread, producing a blank startup and high memory use. Exact precomputed section geometry now lets wrapped reviews mount only nearby files without changing scroll or hunk-navigation coordinates.

## Validation

- `bun run lint`
- `bun run format:check`
- `bun run typecheck`
- `bun run check:docs`
- `bun run changeset:status`
- 96 focused unit/component tests
- 15 targeted PTY tests, including wrapped cross-window hunk navigation and rapid scroll-to-edge recovery
- 64/64 PTY integration tests with an isolated Git/TMP environment
- Compiled binary build
- Frozen 142-file wrapped review (20,853 insertions / 1,480 deletions, 216x60 Kitty), 5/5 valid runs:
  - first paint median 1.403s, max 1.505s
  - peak RSS max 348,618,752 bytes
  - prior affected baseline did not paint within 30s and reached 1,242,595,328 bytes
- Literal local wrapper launch with the compiled candidate: paint, hunk navigation, and clean quit all passed

The broader macOS `bun run test` attempt reached 1,474 passing and 17 skipped tests; three pre-existing session E2E cases remained host-dependent because the harness assumes Linux `script` behavior and has daemon timing sensitivity. The changed surface and PTY integration suites pass, and upstream Linux CI remains authoritative for those session cases.

### DIFF
--- a/.changeset/calm-diffs-window.md
+++ b/.changeset/calm-diffs-window.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Keep large wrapped reviews responsive by rendering nearby files and deferring offscreen highlighting.
--- a/src/ui/components/panes/DiffPane.tsx
+++ b/src/ui/components/panes/DiffPane.tsx
@@ -62,7 +62,6 @@ import { createReviewMouseWheelScrollAcceleration } from "../../lib/scrollAccele
 import {
   buildFileSectionLayouts,
   buildInStreamFileHeaderHeights,
-  collectIntersectingFileSectionIds,
   findHeaderOwningFileSection,
   shouldRenderInStreamFileHeader,
   type FileSectionLayout,
@@ -91,10 +90,11 @@ import { measureFileViewGeometry } from "../../fileViews/geometry";
 import { buildFileViewRenderPlan } from "../../fileViews/renderPlan";
 import { prefetchHighlightedDiff } from "../../diff/useHighlightedDiff";
 import {
-  buildFileRenderWindow,
-  buildFileSectionIndexById,
-  type FileRenderWindowItem,
-} from "../../lib/fileRenderWindow";
+  buildHighlightPrefetchPlan,
+  prefetchImmediateHighlightedFiles,
+  scheduleSpeculativeHighlightedFiles,
+} from "../../diff/highlightPrefetch";
+import { buildFileRenderWindow, buildFileSectionIndexById } from "../../lib/fileRenderWindow";
 import {
   buildCopySelectedRowKeys,
   clampCopyColumn,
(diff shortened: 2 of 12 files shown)


benvinegar (changes requested):
Thanks for tackling this — the performance improvement looks promising, and the focused tests pass. I found two test gaps that I think should be addressed before merging:

1. **The navigation test can hide navigation bugs.**

   `pressUntilVisible` repeatedly sends `]` or `[` until the last or first file appears. That means the test can still pass if a keypress is dropped, a navigation step is duplicated, or one press skips multiple hunks.

   After rebasing, could you use the `ensureKeyboardIsLive` helper now on `main`, then send the expected seven forward and seven backward keypresses explicitly? That would verify each navigation step rather than only checking that the test eventually reaches an edge.

2. **The highlight delay needs timer-focused coverage.**

   The new behavior highlights visible files immediately and waits 300 ms before starting speculative highlighting. The current test only checks that an offscreen file is eventually highlighted, which the previous eager implementation would also pass.

   Could you add controlled-timer tests showing that:
   - selected and visible files start highlighting immediately;
   - speculative files wait for the idle threshold; and
   - scrolling or navigating cancels and restarts pending speculative work?

This branch also conflicts with current `main` in `DiffPane.tsx`, `PierreDiffView.tsx`, and the PTY harness. When resolving those conflicts, please preserve the newer custom file-view and current-line behavior, along with `ensureKeyboardIsLive`.

For validation, I ran the PR head separately: all 96 focused unit/component tests and all 15 navigation/scroll PTY tests passed. Typechecking stopped on the existing unrelated deep-instantiation error in `src/core/customThemes.ts`.

mplibunao:
Thanks for the detailed review. I’ve rebased the branch onto the current `main` and updated the implementation to address the requested changes:

- Navigation now follows the existing keyboard lifecycle through `ensureKeyboardIsLive`, including the exact seven-keypress forward/backward behavior.
- The speculative rendering behavior now has controlled-timer coverage for immediate visible/selected files, the 300 ms delay, cancellation, and timer restart.
- Custom `FileView` behavior and current-line handling are preserved rather than bypassed by the windowing logic.

I also repeated the performance acceptance testing using a compiled build and a real 142-file wrapped diff:

- Real Kitty first-paint median: 839 ms
- Peak Kitty RSS: 367 MiB
- Navigation median: 23 ms
- Five out of five measured runs passed

Compared with current `main`, the updated branch reduced peak memory by roughly 14–15% and improved median first-paint time by about 8%. The full unit suite, lint, typecheck, build, and focused PTY navigation tests also pass. The five failures in the full PTY suite reproduce unchanged on current `main`.

GitHub Actions for the updated fork commit are currently awaiting maintainer approval.

Please let me know if you’d prefer any further changes to the windowing or timer behavior.

PR #621, opened 2026-07-28
fix(nix): keep flake evaluable on nixpkgs without x86_64-darwin

### LINKED ISSUE #608: `hunk`/`bun2nix` flake evaluation fails on nixpkgs 26.11+ due to `x86_64-darwin` in supported systems
### What happened?

Consuming `hunk` as a flake input on `nixpkgs-unstable` (specifically nixpkgs 26.11+) fails during evaluation with:

```
error: Nixpkgs 26.11 has dropped support for x86_64-darwin.

The 26.05 stable branch still supports x86_64-darwin, and will
receive security fixes until the end of 2026. ...
```

This happens even when the consumer only ever accesses the `aarch64-darwin` output (e.g. `inputs.hunk.packages.${pkgs.stdenv.hostPlatform.system}.hunk` on Apple Silicon). The failure occurs while `flake-parts` is evaluating/transposing `perSystem` outputs across all declared systems, before the consumer's per-system accessor ever narrows the selection down.

### Steps to reproduce
1. Use a flake with:
   ```nix
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
     hunk = {
       url = "github:modem-dev/hunk";
       inputs.nixpkgs.follows = "nixpkgs";
     };
   };
   ```
2. Reference `inputs.hunk.packages.${pkgs.stdenv.hostPlatform.system}.hunk` in `environment.systemPackages` (nix-darwin) on an `aarch64-darwin` host.
3. Run `nix build` / `darwin-rebuild switch` once `nixpkgs-unstable` has advanced to the 26.11 release that drops `x86_64-darwin`.
4. Observe the evaluation error, with a trace through `flake-parts/modules/transposition.nix` and `bun2nix`'s `flake.nix`.

#### Environment
- `nixpkgs`: `nixpkgs-unstable` (26.11-era, dropped `x86_64-darwin`)
- `nix-darwin` host system: `aarch64-darwin`
- `hunk` flake input: `github:modem-dev/hunk` (no `ref` pinned, tracks default branch)
- Consumption method: flake input in a nix-darwin `environment.systemPackages`, exactly as documented in hunk's README ("Install from a flake")

#### Workaround applied locally
Removed the `hunk` flake input and its package reference from `environment.systemPackages`; installed `hunk` via nix managed Homebrew instance. 

### Expected behavior

Consuming `hunk` on an `aarch64-darwin` (Apple Silicon) host with current nixpkgs-unstable should not require `x86_64-darwin` to evaluate/build successfully.

## Suggested fix
- Drop `x86_64-darwin` (or Intel macOS entirely) from `hunk`'s `supportedSystems` list, since Nixpkgs itself no longer supports it as of 26.11.
- Update/pin `bun2nix`'s `systems` input away from `nix-systems/default` (which still includes `x86_64-darwin`) to a subset that excludes it, or make it configurable via `inputs.systems.follows` so downstream consumers can restrict it without forking.

### Version

0.17.6

elucid:
Fixes #608

## Problem

Nixpkgs 26.11 dropped `x86_64-darwin` and now *throws* on `import nixpkgs { system = "x86_64-darwin"; }`. Consuming Hunk as a flake input on `nixpkgs-unstable` fails to evaluate on Apple Silicon, even though only the `aarch64-darwin` output is ever referenced.

There are two independent forcing paths, and only fixing the first one is not enough:

1. Hunk's `supportedSystems` declared `x86_64-darwin`. This is lazy, so it only broke `nix flake check --all-systems` (which our own Nix CI job runs).
2. **The one that actually breaks consumers:** bun2nix's derivations reference flake-parts module args (`self'`) for *every* system in bun2nix's `systems` input, which was pinned to `nix-systems/default` (includes `x86_64-darwin`). So `packages.aarch64-darwin.hunk.drvPath` still throws with the reported trace through `flake-parts/modules/transposition.nix`, even after dropping the system from our own list.

## Change

Add a `systems` flake input as the single source of truth for the systems Hunk and bun2nix are evaluated for:

```nix
systems.url = "github:nix-systems/triplet";   # x86_64-linux, aarch64-linux, aarch64-darwin
bun2nix.inputs.systems.follows = "systems";
...
supportedSystems = import systems;
```

Consumers who pin an older Nixpkgs that still supports a dropped system can widen the list with `inputs.hunk.inputs.systems.follows`, documented in `nix/README.md`.

Intel macOS users should install Hunk via npm or Homebrew; Nixpkgs no longer supports that platform on unstable.

`meta.platforms` in `nix/package.nix` intentionally still lists `x86_64-darwin` — it describes what Hunk supports (npm/Homebrew ship Intel builds) and keeps the `systems.follows` override path working on 26.05.

## Verification

On `aarch64-darwin`, Nix 2.31:

| check | before | after |
| --- | --- | --- |
| `nix eval .#packages.aarch64-darwin.hunk.drvPath` with `--override-input nixpkgs github:NixOS/nixpkgs/nixpkgs-unstable` | ❌ 26.11 throw | ✅ |
| `nix flake check --all-systems --no-build` with unstable override | ❌ | ✅ |
| `nix flake check --all-systems --no-build` with the committed lock | ✅ | ✅ |
| `nix build .#hunk` (binary runs, reports `0.17.0`) | ✅ | ✅ |

### DIFF
--- a/.changeset/nix-flake-drop-x86-darwin.md
+++ b/.changeset/nix-flake-drop-x86-darwin.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Fix Nix flake evaluation on Nixpkgs 26.11, which dropped `x86_64-darwin`. Hunk's flake no longer declares that system, and it now pins bun2nix's `systems` input to the same list so building the `aarch64-darwin` package never forces an Intel macOS Nixpkgs. The system list is exposed as a `systems` flake input for consumers that need to override it.
--- a/flake.lock
+++ b/flake.lock
@@ -7,7 +7,9 @@
         "nixpkgs": [
           "nixpkgs"
         ],
-        "systems": "systems",
+        "systems": [
+          "systems"
+        ],
         "treefmt-nix": "treefmt-nix"
       },
       "locked": {
@@ -91,21 +93,22 @@
     "root": {
       "inputs": {
         "bun2nix": "bun2nix",
-        "nixpkgs": "nixpkgs"
+        "nixpkgs": "nixpkgs",
+        "systems": "systems"
       }
     },
     "systems": {
       "locked": {
-        "lastModified": 1681028828,
-        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
+        "lastModified": 1776166891,
+        "narHash": "sha256-bI8yrEGjrohR5hkQox7UrxDH7XqrYMwI8SL/LrJ1+S8=",
         "owner": "nix-systems",
-        "repo": "default",
-        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
+        "repo": "triplet",
+        "rev": "6de7bc09397911ce03636afbcf6118745ab2cda0",
         "type": "github"
       },
       "original": {
         "owner": "nix-systems",
-        "repo": "default",
+        "repo": "triplet",
         "type": "github"
       }
     },
--- a/flake.nix
+++ b/flake.nix
@@ -2,8 +2,18 @@
   description = "Dev Flake for Modem-dev's Hunk";
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    # Systems Hunk builds for. Nixpkgs 26.11 dropped x86_64-darwin, and
+    # importing nixpkgs for a dropped system throws, so every declared system
+    # has to stay evaluable. Consumers can widen or narrow this with
+    # `inputs.hunk.inputs.systems.follows`.
+    systems.url = "github:nix-systems/triplet";
     bun2nix.url = "github:nix-community/bun2nix";
     bun2nix.inputs.nixpkgs.follows = "nixpkgs";
+    # bun2nix builds its flake-parts outputs for every system in its `systems`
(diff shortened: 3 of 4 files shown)


elucid:
Addressed the Greptile note on the `nix/README.md` override example — it was a real bug, not a nit.

Verified the original example fails:

```
error: input 'hunk/bun2nix/systems' follows a non-existent input 'hunk/systems'
```

The error points at bun2nix, which would be a confusing thing to hit while following the doc.

Rather than the suggested root-input + `follows` form, the example now uses the self-contained override, verified end-to-end against a Nixpkgs that still has `x86_64-darwin` (pinned `13043924`, 2026-04-11):

```nix
inputs.hunk.inputs.systems.url = "github:nix-systems/default"; # adds x86_64-darwin
```

That resolves `hunk.packages.x86_64-darwin.hunk` to a real derivation, confirming the single `systems` input widens both Hunk and bun2nix. The `follows` form is mentioned as an alternative for flakes that already declare their own `systems` input, and it was verified too.

Docs-only change; `flake.nix`/`flake.lock` are untouched. Re-ran `nix flake check --all-systems --no-build` (committed lock), the `aarch64-darwin` drvPath eval against `nixpkgs-unstable`, and the clean cherry-pick onto `0.17.x` — all still pass.

PR #633, opened 2026-07-28
feat(extensions): add setTheme method to extension API

gogamid:


benvinegar:
Can I ask why / what you were trying to do w/ this? (Not that it isn't valid, I'd just like to understand)

gogamid:
@benvinegar to be able to change the theme live for an already open hunk diff when I change macos appearance between dark and light.

PR #647, opened 2026-07-30
feat(pager): give pager mode the full review controls

### LINKED ISSUE #550: pager does not support operations that diff does (e.g. previous/next file)
### What happened?

`git diff origin/main... | hunk pager` seems nerfed compared to `hunk diff origin/main...`

- file-based navigation does not work (even though sidebar looks correct)
- indeed the entire menu is unavailable
- `j`/`k` jump multiple lines instead of one at a time in `hunk diff` (separate bug?)
- `z` doesn’t work
- changing view doesn’t work

My use case is I am filtering the diff with `grep -v` which is why I’m not just using `hunk diff`

### Steps to reproduce

Pipe a diff into `hunk pager` and observe that you cannot navigate by file or do other `hunk diff` things like pick a different view

### Version

0.17.1

HackAttack:
Pager mode was doing two unrelated jobs under one flag. Besides choosing the chrome, `options.pager` gated capability: every command carried a scope tag and only a dozen scrolling commands were tagged for pager, so file navigation, layout switching, the menu, help, and the theme selector were filtered out before they could run. Render guards in App then suppressed the surfaces those commands would have drawn.

Dropping the scope split and the guards preserves the chrome preset without nerfing the functionality. File navigation, layout, etc. now work in pager mode.

Fixes #550


elucid:
@HackAttack thanks for this! This is the better approach. I've switched my git pager now.

Issue #514, opened 2026-07-06
Can we have a config setting for displaying the sidebar?


ibrahimn-flwls:
### What do you want to change?

It would be nice if we had a setting to display/turn off the sidebar by default.

### Why?

I like having the sidebar open and seeing all the files changed.

nikok6:
Agreed. The automatic sidebar only triggers at 220 columns wide minimum. Would be nice if there is a config setting.

skaragianis:
Addressed by https://github.com/modem-dev/hunk/pull/648 and available in https://github.com/modem-dev/hunk/releases/tag/v0.19.0, thanks @benvinegar

benvinegar:
Closing as completed by #648. Hunk now supports `sidebar = "auto" | true | false` in config, along with `--sidebar` and `--no-sidebar`, and the change shipped in v0.19.0.

*This comment was generated by Pi using gpt-5.6-sol*

PR #648, opened 2026-07-31
feat(ui): add configurable sidebar display

skaragianis:
I often turn on the sidebar to keep an eye on agent activity and scope. This makes it a configuration.

Adds `--sidebar` and `--no-sidebar` CLI flags, and a `sidebar` configuration with options `"auto"`, `true`, and `false`. Default remains `auto`. `true` will not override pager mode or the responsive layout. `false` will override the responsive layout. An 's' keypress continues to toggle where the responsive layout allows.

Addresses ([#514](https://github.com/modem-dev/hunk/issues/514))


skaragianis:
`bun check:docs` now passes on https://github.com/modem-dev/hunk/commit/7b1b2a4244a51fe68d5d06dbb1c34e9f39b1b7ae, if you could approve the pending runs on https://github.com/modem-dev/hunk/pull/648/commits/8294f571a1a80100350b556b10dea40a3d756368 please.

skaragianis:
I rebased on `main` to resolve the conflict with #675 and cleaned up the formatter churn I accidentally left in `App.tsx` (which is what caused the conflict). No change to the feature itself - the diff is the same minus the whitespace. If you could kick it along again please?

skaragianis:
@benvinegar will you be swinging past again?

benvinegar (changes requested):
Thanks for the follow-up. The feature looks sound on its original base, and the current head passes typecheck, the targeted tests, and the generated-docs check. Two things need addressing before merge:

1. Please rebase onto current `main`. #757 changed `s` to toggle only the active files pane, while this PR's `sidebarState` still filters every left/right pane through `effectiveOpenPaneKeys`. Resolving the `App.tsx` conflict mechanically would regress independently opened extension panes. Please apply the configured initial state to the files-pane slot/open state while preserving other extension panes, with coverage for the combined behavior.
2. `buildCommonOptions` scans the full raw `argv`, including pathspecs after `--`. As a result, `hunk diff -- --sidebar` and `hunk diff -- --no-sidebar` both keep the pathspec *and* change sidebar state. Please resolve these flags only from command-option tokens before the separator and add a regression test.

benvinegar (approved):
Rebased onto current `main` and addressed both findings. The sidebar preference now controls only the active files-pane slot while preserving independent extension panes, including replacements. Paired boolean flags also stop at the `--` separator, with table-driven regression coverage.

jesseleite:
Just wanted to say thank you both for this one! 🎉

PR #649, opened 2026-07-31
fix(ui): route keys by ownership so modal surfaces stop double-handling them

elucid:
## What was broken

OpenTUI delivers every keypress to Hunk's **global handler chain first** and to the **focused renderable last** — the inverse of the browser model. So when a global handler "declines" a key, it isn't leaving the key alone: it passes it along the chain and ultimately to a focused widget that may act on it too. OpenTUI scroll boxes answer arrows and `j`/`k` with a fifth-of-a-viewport scroll of their own; text inputs insert characters.

Our handlers returned a boolean, which can only say *"stop asking other handlers"* — it cannot say *"and stop the key"*. Several handlers acted on keys without consuming them, so one keypress triggered two actions:

- **Pager mode, menu open, press an arrow** — the menu selection moves *and* the diff scrolls **21 rows** behind the dialog. Same with any key the theme selector means to swallow.
- **Escape** closing the Controls help (or Agent skill) overlay *also* wiped typed filter text — silent loss of user-typed input.
- **Enter** activating a menu item *also* submitted the focused filter.
- **F10** popped the menu bar over an in-progress note draft.

The pager cases became user-reachable in #647: pager focuses the review scroll box by design (`focused={pagerMode}`), and #647 gave pager the full review controls, so menus and the theme selector now open over a focused scroll box. Verified against `aa1c24a` — 21 rows of phantom scroll in both.

#645 fixed one instance of this class (command dispatch not consuming). This PR fixes the rest and makes the mistake unrepresentable.

## The model

Every handler now answers one question — **who owns this key?** — with a three-state `KeyOwner` (`src/ui/lib/keyRouting.ts`):

| answer | ends chain? | widget receives key? | consumed? |
| --- | --- | --- | --- |
| `"notMine"` | no | decided later | decided later |
| `"mine"` | yes | **no** | yes — centrally, by the dispatch loop |
| `"focused"` | yes | **yes** | no |

Why three states: a keypress raises two independent questions — *stop the chain?* and *stop the key?* A boolean answers only the first. The fourth combination (keep asking other handlers but suppress the widget) is incoherent, so three values are exactly enough.

Consumption is enforced centrally in `routeKeyOwnership`: a handler can no longer act on a key and forget to `preventDefault()` it, because "acted" and "consumed" are the same return value.

**For future authors:** the everyday decision is two-state — did your handler act on (or deliberately swallow) the key? `"mine"`. Otherwise `"notMine"`. `"focused"` has exactly one trigger: *a focused text input needs this key as text* (the file filter, the inline note draft, an extension input dialog). If you're about to write a comment like "let the input own this", you want `"focused"`. Unless you're adding a new text input, you will never return it.

Two fallthroughs are deliberate and covered by tests: menu accelerators (`q`, `?`, `/`…) still pass an open menu to the command table, and step keys still scroll exactly one row with a menu open — an open menu is intentionally not fully modal.

## Also in this PR

- Deletes `AgentInlineNote`'s duplicated Escape/Ctrl-S textarea handling. The chain now provably consumes both before the textarea can see them, so the copy was unreachable duplicated logic.
- Lifts `measureKeyScroll` (added in #645) from `test/pty/scroll.test.ts` into the shared PTY harness.

## Tests (written red-first)

- `src/ui/lib/keyRouting.test.ts` — the ownership contract: `notMine` walks on, `mine` consumes and stops, `focused` stops without consuming.
- `test/pty/key-routing.test.ts` — end-to-end in a real PTY: pager menu arrows and theme-selector keys must not scroll the stream (red on `aa1c24a` at 21 rows each); one Escape closes the overlay and keeps filter text; Enter runs the menu item without submitting the filter; F10 stays dead over a note draft; plus the two deliberate-fallthrough guards.
- `src/ui/AppHost.key-routing.test.tsx` — the same two scroll-box cases in review mode, with focus handed to the scroll box directly.

## Verification

- Rebased onto `aa1c24a`. The conflict was semantic, not textual: #647 removed `pagerMode` from this hook (pager has a menu bar now) and dropped the scope argument from `dispatchAppCommand`. Resolved by deleting my `pagerMode` F10 branch entirely while keeping the note-draft guard, and adopting the new dispatch signature.
- `bun run typecheck`, `bun run lint`, `bun run format` — clean.
- `bun run test` / `bun run test:integration` — failure sets identical to a clean worktree at `aa1c24a` (network-dependent update-notice tests, daemon-dependent session tests, and 7 pager PTY tests that time out on piped stdin on this machine); no regressions.

### DIFF
--- a/.changeset/key-owner-routing.md
+++ b/.changeset/key-owner-routing.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+One keypress no longer triggers two actions when a modal surface and the focused widget both want it. In pager mode, opening a menu or the theme selector and pressing an arrow no longer scrolls the diff behind the dialog. Escape closing the Controls help or Agent skill overlay no longer wipes typed filter text, Enter on a menu item no longer also submits the focused filter, and F10 no longer pops the menu bar over an in-progress note draft. Global key handlers now answer one three-state ownership question (`notMine` / `mine` / `focused`) and consumption is enforced centrally, so a handler can no longer act on a key while leaving it live for the focused widget.
--- a/src/ui/AppHost.key-routing.test.tsx
+++ b/src/ui/AppHost.key-routing.test.tsx
@@ -1,0 +1,192 @@
+import { describe, expect, mock, test } from "bun:test";
+import { ScrollBoxRenderable, type Renderable } from "@opentui/core";
+import { testRender } from "@opentui/react/test-utils";
+import { act } from "react";
+import type { AppBootstrap } from "../core/types";
+import { createTestVcsAppBootstrap } from "../../test/helpers/app-bootstrap";
+import { createTestDiffFile } from "../../test/helpers/diff-helpers";
+
+mock.restore();
+
+const { AppHost } = await import("./AppHost");
+
+/**
+ * Key-ownership routing between the global handler chain and a focused
+ * scroll box.
+ *
+ * OpenTUI runs Hunk's global key handler before the focused renderable, and
+ * its scroll boxes answer arrows and j/k with a fifth-of-a-viewport scroll of
+ * their own. Any modal surface that acts on a key without consuming it
+ * therefore scrolls the review stream behind itself whenever the scroll box
+ * holds focus — which it always does in pager mode, and does in review mode
+ * once focus lands on it. These tests pin the modal surfaces that must own
+ * their keys outright.
+ */
+function createScrollableBootstrap(): AppBootstrap {
+  // Change every line so no context collapses: the review stream must be
+  // taller than the viewport for the scroll box to have anywhere to go.
+  const before = Array.from(
+    { length: 80 },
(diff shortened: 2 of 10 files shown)


elucid:
Examples of what this fixes:

1. escape no longer eats filter text
  `? → / → type "beta" → Escape (once)`

 - New: help overlay closes, status bar still reads filter: beta
 - Old: overlay closes and your text is gone - status bar resets to type to filter files

2. Enter on a menu item no longer submits the filter
  `/ → type beta → F10 → → → ↓ → Enter (View ▸ Stacked view)`

 - New: layout goes stacked, status bar still reads filter: beta (focused)
 - Old: layout goes stacked and the filter submits. note: filter=beta (blurred)
   instead of filter: beta

3. F10 no longer pops the menu over a note draft
  `c (start a note) → F10 → keep typing`

 - New: nothing happens, draft keeps focus, typing flows into the note
 - Old: the File menu opens on top of your in-progress draft

PR #662, opened 2026-08-01
feat(review): mark the current line and anchor notes to it

### LINKED ISSUE #436: Line-level focus for better keyboard-driven user comments

Add keyboard-accessible line focus tracking within the scrolling diff view, so it can be used as a target for creation a comment with the `c` keybind.

### Why?

So Hunk already has line-level scrolling with j/k, but it only scrolls the viewport and does not come with a line focus. This causes problems in a keyboard-driven workflow when I e.g. want to create a comment on a particular line. as is, the `c` keybind just creates it "somewhere" in the hunk, and not necessarily the line i want to draw an agent's focus to.

This also brings parity to the mouse-driven workflow, where you CAN click a specific line and hit the `[+]` to leave a line-level comment.

### How? (optional)

_No response_

### LINKED ISSUE #553: Current line is not very apparent
### What happened?

It is not at all clear what line I am hovering in the `hunk diff` UI. I want to add a comment at a specific line, but can't.

### Steps to reproduce

Open `hunk diff`, and see no visible indicator of what line you are on.

### Expected behavior

See a visual indicator of current line.

### Version

0.17.0

loganthomas:
Closes #553. Closes #436.

## Behavior

- The line you are on is highlighted, and `j`/`k` (and `↑`/`↓`) move it. The view scrolls only far enough to keep it visible.
- `c` anchors a note at that line instead of the top of the hunk. A live mouse hover still wins, so the pointer flow is unchanged.
- The cursor rolls across hunk and file boundaries, clamps at both ends, and stays in step with `[`/`]` and sidebar selection in both directions.
- `cursor_line` = `row` (default, whole row) | `number` (line-number gutter only) | `off`. Settable in config, with `--cursor-line <style>`, or from the View menu.
- **`off` restores today's `j`/`k` scrolling exactly.**

## On `j`/`k`

`AGENTS.md` says not to reintroduce `j`/`k` *hunk* navigation. This is line navigation, which is what #436 asks for; one commenter there cannot use a mouse. `off` opts out, and the PTY guards that asserted one-row step scrolling now run in that mode, keeping their original coverage.

Say the word and I'll flip the default to `off`.

## Implementation

Hunk already resolves a line-precise note target; it just expires ~2s after the pointer moves. This makes it persistent and keyboard-driven.

- `src/ui/lib/lineCursors.ts` mirrors `hunks.ts` one level down, memoized per parsed file so a keypress does not reallocate the list.
- `useReviewController` owns the cursor beside the existing selection, minting a reveal id like `selectHunk` does.
- Rendering resolves one `RowHighlight` per cell (blend plus optional column span) and hands it to the existing cell renderers, replacing the old `selected` + `selectionColRange` pair. No new theme tokens. The cursor row is found by the render plan's own `line:<hunk>:<side>:<n>` key, so windowing needs no special case.
- On split rows only the cursor's side is highlighted, since the two halves are different note targets.

The marker shifts **luminance**, not hue: blending toward one fixed highlight color barely moves a background that already shares its hue, which left it invisible on added rows. `AppHost.cursor-line.test.tsx` asserts the painted backgrounds, including that added stays green and removed stays red under the marker.

## Known gaps

Targets come from `hunkContent` rather than the render plan, so lines revealed by expanding a collapsed gap are not reachable by `j`/`k`. `z` still expands the gap at the cursor's hunk. Same for files shown through an extension file presentation, which have no `line:` anchors. Both are called out in the module docstring; happy to follow up with a plan-derived target list if you'd prefer that before merge.

## Verification

`format:check`, `lint`, `typecheck`, `check:docs`, `test`, `test:integration`, `test:tty-smoke`, `build:npm`, `check:pack`. Docs updated: keybindings table, keyboard reference, notes guide, display config, plus regenerated CLI/config reference.

Pre-existing failures on `main` at `e0d1757`, unchanged by this PR (verified by stashing it): `AppHost.watch.test.tsx` theme resolution, and 5 PTY extension/file-view tests.

No true interactive TTY run: no controlling terminal here. Instead the real painted spans are asserted via `captureSpans()` across `row`/`number`/`off`, split and stack, light and dark.

### DIFF
--- a/.changeset/current-line-cursor.md
+++ b/.changeset/current-line-cursor.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": minor
+---
+
+Highlight the current line and move it with `j`/`k`, so `c` adds a note exactly where the cursor sits. Set `cursor_line` to `number` for a quieter line-number marker, or `off` to restore plain row scrolling.
--- a/docs/keybindings.md
+++ b/docs/keybindings.md
@@ -78,6 +78,9 @@ The built-in commands and the keys they ship with:
 | `hunk.view.layoutSplit`             | Split layout                             | `1`                          |
 | `hunk.view.layoutStack`             | Stack layout                             | `2`                          |
 | `hunk.view.layoutAuto`              | Auto layout                              | `0`                          |
+| `hunk.view.cursorLineRow`           | Highlight the current row                | _(none)_                     |
+| `hunk.view.cursorLineNumber`        | Mark the current line number             | _(none)_                     |
+| `hunk.view.cursorLineOff`           | Hide the current-line marker             | _(none)_                     |
 
 Commands marked _(none)_ ship without a key: they are menu items today, and
 binding one gives it a shortcut like any other.
--- a/src/core/cli.test.ts
+++ b/src/core/cli.test.ts
@@ -151,6 +151,19 @@ describe("parseCli", () => {
     });
   });
 
+  test("parses the current-line style and rejects an unknown one", async () => {
+    const parsed = await parseCli(["bun", "hunk", "diff", "--cursor-line", "number"]);
+
+    expect(parsed).toMatchObject({
+      kind: "vcs",
+      options: { cursorLine: "number" },
+    });
+
+    await expect(parseCli(["bun", "hunk", "diff", "--cursor-line", "sparkles"])).rejects.toThrow(
+      "Invalid cursor line style: sparkles",
+    );
+  });
+
   test("accepts --experimental before the review command", async () => {
     const parsed = await parseCli(["bun", "hunk", "--experimental", "diff"]);
 
--- a/src/core/cli.ts
+++ b/src/core/cli.ts
@@ -4,6 +4,7 @@ import { Command, Option } from "commander";
 import type {
   CliInput,
   CommonOptions,
+  CursorLine,
   HelpCommandInput,
   LayoutMode,
   PagerCommandInput,
@@ -38,7 +39,7 @@ import { resolveCliVersion } from "./version";
 export interface CliReferenceOption {
   readonly flag: string;
   readonly description: string;
-  readonly parse?: "layout" | "positiveInt" | "tabWidth" | "collect";
+  readonly parse?: "layout" | "cursorLine" | "positiveInt" | "tabWidth" | "collect";
   readonly defaultValue?: string;
   /** Default applied directly by Commander (as opposed to a config-resolved default). */
   readonly commanderDefault?: string;
@@ -59,6 +60,11 @@ export interface CliReferenceCommand {
 /** Review flags registered on every full-screen review command. */
 export const COMMON_REVIEW_OPTIONS = [
   { flag: "--mode <mode>", description: "layout mode: auto, split, stack", parse: "layout" },
+  {
+    flag: "--cursor-line <style>",
+    description: "current-line marker: row, number, off",
+    parse: "cursorLine",
+  },
   { flag: "--theme <theme>", description: "named theme override" },
   AUXILIARY_AGENT_OPTIONS.agentContext,
   { flag: "--pager", description: "use pager-style chrome" },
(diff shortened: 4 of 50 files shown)


loganthomas:
## Examples
`bun run src/main.tsx -- show HEAD --cursor-line row`
Screenshot 2026-08-01 at 17 22 32
Screenshot 2026-08-01 at 17 22 41

`bun run src/main.tsx -- show HEAD --cursor-line number`
Screenshot 2026-08-01 at 17 22 57
Screenshot 2026-08-01 at 17 23 06

loganthomas:
Added to menu bar as well (toggle without having to pass command line args)

Screenshot 2026-08-01 at 17 27 13

benvinegar:
@loganthomas Dude, figuring out the UI for this has been my biggest struggle - I love how high quality this PR is, that you explored a bunch of different options, etc.

benvinegar:
This was AI generated.

I checked this out and exercised it in tmux. The precise note anchoring and `cursor_line = "off"` fallback work, but I found a few regressions:

- Transparent mode passes `"transparent"` into the hex blender, so `github-light-default --transparent-bg` paints the current context row nearly black (`#060708` under `#1f2328` text, ~1.28:1 contrast).
- Paging or wheel-scrolling can leave the cursor off-screen. Pressing `c` then anchors at that stale line while preserving the viewport, leaving no visible draft editor.
- Cursor order diverges from rendered rows: split replacements traverse every old line before jumping back to the first new line, and expanded context is skipped entirely. In tmux, an expanded gap jumped from line 13 to line 107 on one `j`.
- Alternate file presentations do not receive cursor rendering or compatible reveal geometry, so default `j`/`k` updates an invisible raw-diff cursor instead of highlighting or scrolling the active presentation.
- On a selected file with no navigable lines, `firstLineCursorInHunk` falls back to another file's first cursor; `c` can therefore open a note on an unrelated file.

I think the navigation model should come from the same rendered plan/geometry used for display and note placement, with note creation always revealing its target.

*This comment was generated by Pi using OpenAI GPT-5.3 Codex*

loganthomas:
@benvinegar sound good! I'll pick this up Monday. Excited to help on this cool project, thanks for building!

benvinegar:
NP!

> Paging or wheel-scrolling can leave the cursor off-screen. 

Aside, I'm thinking if you navigate by page we just have to choose a new line somewhere (e.g. middle of viewport).

There's a bunch of UX considerations like this. (This is why I was scared to tackle it, aha.)

loganthomas:
Thanks for digging into this, the tmux pass was really helpful.

Fixed:
- Paging and wheel scrolling leaving the cursor off screen. It stays put while it's visible now, and snaps to the nearest visible line once you scroll past it.
- `c` opening with no visible editor. You were right it was broken, though the cause turned out to be different. The reveal was scrolling to the hunk's first note instead of the draft, and the draft's viewport restore was overriding it anyway.
- Expanded context being skipped.
- Alternate file presentations getting no cursor or reveal geometry.

You were right about the root cause too. Navigation now comes from the measured render plan instead of a separate walk over hunkContent, so the ordering problems can't really come back.

Three I couldn't reproduce on the current branch:
- Transparent mode measures 10.54:1 (#d2d3d4 under #1f2328). #060708 is what you get without the sentinel guard, and that guard is there.
- Split already pairs old and new per row.
- firstLineCursorInHunk stays inside the file and returns null.

I think the branch got rewritten after your review ran, so you may have been looking at an earlier push. Not trying to wave them off.

On your paging note, I went with snapping to the nearest visible line instead of the middle of the viewport. It still picks a line every page, and it has the nice side effect of leaving the marker alone when you nudge the wheel one line.

One other change: note cards now show up below the line they're attached to instead of above.

And a question, since you mentioned there are a lot of UX calls like this. Expanding a gap pushes the hunk and the marker off the bottom of the screen. That happens on main too, but the marker makes it obvious. Right now `z` moves the marker to the first revealed line and scrolls there, then puts it back when you collapse. The other option is to leave the marker where it is and hold your place, so the revealed lines scroll in above you. I genuinely don't know which feels better. What do you think?

benvinegar:
> Right now z moves the marker to the first revealed line and scrolls there, then puts it back when you collapse. The other option is to leave the marker where it is and hold your place, so the revealed lines scroll in above you. I genuinely don't know which feels better. What do you think?

I think it's less confusing if the active line stays where it is (vs jumping).

Or let's just start with that and see if it needs to change after?

benvinegar:
Actually let's just keep it the way you have it. If it's irritating we'll learn and find out. LGTM!

PR #676, opened 2026-08-04
feat(review): select current-line sides and add named sidebar policies

rupert648:
## Summary

- adds `h` / `l` keyboard selection for the old and new sides of the current split row
- makes `c` anchor an exact line comment on the selected side
- defines sidebar policy as `auto`, `shown`, or `hidden`, with boolean config values retained as compatibility aliases
- documents and tests the new keyboard and configuration behavior

## Stacked dependencies

This draft intentionally preserves and integrates the commits from #662 and #648 so the combined behavior can be used and tested now. Original authorship is preserved. The new delta is commit `0d27fa8`.

This should be rebased after those pull requests land so the final review contains only the side-selection and named-policy delta.

## Validation

- format, lint, typecheck, and generated-doc checks pass
- focused UI, command, config, cursor, and sidebar tests pass
- npm build and package validation pass
- filesystem watcher and daemon lifecycle tests pass with loopback access
- macOS TTY smoke is blocked by the repository test harness using GNU `script -f`, which Apple `script` does not support


benvinegar:
Could this be done with an extension?

I feel like I need help understanding the goals.

PR #678, opened 2026-08-04
fix(cli): accept `hunk diff A B` as the two-commit range A..B

HackAttack:
Git treats `diff A B` and `diff A..B` as the same two-commit review, but Hunk only understood the dotted spelling: a second positional was always folded into pathspecs, so `hunk diff main feature` silently reviewed `main` limited to a pathspec named `feature`.

Two positionals now normalize to `A..B`, except where the old reading is the intended one: a first target that already spells a range, or a second token that exists on disk, which is what `diff <rev> <path>` without a `--` separator relies on.

### DIFF
--- a/.changeset/diff-two-commit-args.md
+++ b/.changeset/diff-two-commit-args.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Accept `hunk diff A B` as the same two-commit review as `hunk diff A..B`.
--- a/.changeset/diff-two-commit-args.md
+++ b/.changeset/diff-two-commit-args.md
@@ -1,5 +1,5 @@
 ---
-"hunkdiff": patch
+"hunkdiff": minor
 ---
 
-Accept `hunk diff A B` as the same two-commit review as `hunk diff A..B`.
+Accept `hunk diff A B` as a two-commit review, same as Git's `A..B`. A pathspec following a single target now needs a `--` separator.
--- a/src/app/cli.test.ts
+++ b/src/app/cli.test.ts
@@ -391,47 +391,67 @@ describe("parseCli", () => {
     });
   });
 
-  test("treats two revision positionals as an A..B range", async () => {
+  test("treats two revision positionals as the two commits to compare", async () => {
     const parsed = await parseCli(["bun", "hunk", "diff", "main", "feature"]);
 
     expect(parsed).toMatchObject({
       kind: "vcs",
-      range: "main..feature",
+      rangeEndpoints: { from: "main", to: "feature" },
       staged: false,
     });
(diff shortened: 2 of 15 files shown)


src/core/cli.ts
```diff
@@ -649,16 +656,34 @@ async function parseDiffCommand(tokens: string[], argv: string[]): Promise<Parse
...
       };
     }
 
+    // Git reads `diff A B` as `diff A..B`, so Hunk should too. A target that already
+    // spells a range keeps its trailing token as a pathspec, and so does a second token
+    // that exists on disk, since that is the shape `diff <rev> <path>` relies on.
+    if (!isRangeExpression(left) && (normalizedPathspecs !== undefined || !existsSync(right))) {
+      return {
+        kind: "vcs",
+        range: `${left}..${right}`,
```
benvinegar: **This breaks two-commit diffs in jj and Sapling.**

`A..B` is Git syntax, but this conversion happens before VCS detection. In jj, `jj diff -r A..B` does not always compare A to B; with diverged commits it can omit A-side removals. Keep the endpoints separate and let each adapter translate them.

*This comment was generated by Pi using OpenAI Codex gpt-5.6-sol*

src/core/cli.ts
```diff
@@ -649,16 +656,34 @@ async function parseDiffCommand(tokens: string[], argv: string[]): Promise<Parse
...
+        left,
+        right,
         options,
       };
     }
 
+    // Git reads `diff A B` as `diff A..B`, so Hunk should too. A target that already
+    // spells a range keeps its trailing token as a pathspec, and so does a second token
+    // that exists on disk, since that is the shape `diff <rev> <path>` relies on.
+    if (!isRangeExpression(left) && (normalizedPathspecs !== undefined || !existsSync(right))) {
```
benvinegar: **This depends too much on what happens to exist on disk.**

If a branch and directory are both named `feature`, `hunk diff main feature` becomes a path-limited diff. Missing, deleted, or glob pathspecs go the other way and become invalid revision ranges. Please use VCS-aware revision detection, or require `--` for pathspecs, instead of `existsSync`.

*This comment was generated by Pi using OpenAI Codex gpt-5.6-sol*

benvinegar (changes requested):
Two argument-parsing issues need to be fixed before merge.

*This comment was generated by Pi using OpenAI Codex gpt-5.6-sol*

HackAttack:
Both true, fixed!

HackAttack:
Superseded by #938

benvinegar:
@HackAttack Ah, sorry, there was so much thrash around these changes, that I re-built everything from your work. This really should have your attribution; I'm sorry about that.

HackAttack:
No worries at all! Thanks for integrating.

PR #682, opened 2026-08-07
fix(ui): wrap draft notes by terminal cells so long CJK input stays visible

### LINKED ISSUE #681: Draft note composer never wraps CJK text — one-line textarea scrolls horizontally
### What happened?

When typing a review note in the draft composer (press `c` on a hunk), CJK text never soft-wraps at the right moment. The composer stays one row high and only shows the visual row the cursor is on, so previously typed text becomes invisible while typing. 
Right before the wrap threshold the box looks full but is still one row
(figure 1):
Image

After typing one more character, the view jumps to the overflow visual row and the composer appears to "lose" everything except the last typed character (figure 2). The text is not lost — the 1-row-high viewport
follows the cursor and hides the rest:
(figure 2):
Image

An English note of the same display length wraps and grows rows correctly.

Root cause — two compounding issues:

1. `draftVisualLineCount` in `src/ui/components/panes/AgentInlineNote.tsx` estimates the composer height with `String#length` (UTF-16 code units) instead of terminal cells — the same bug class as #566, in a code path that #567 didn't touch. A 60-char CJK string (120 cells) in a ~76-cell box estimates to `ceil(60/76) = 1` row, so the composer height stays 1 until the *character count* exceeds the box width (for CJK that means roughly two full visual rows of text).
2. OpenTUI's `TextareaRenderable` reports `virtualLineCount` capped at the viewport height: a 1-row-high textarea never starts wrapping, so the `max(estimate, virtualLineCount)` correction in `onContentChange` can never grow the height either — a deadlock. The editor keeps the cursor visible by scrolling the capped viewport vertically, which is exactly what produces figure 2.

The deadlock only breaks when the character count finally exceeds the box width, the estimate grows to 2, and wrapping suddenly kicks in — which is why the composer *sometimes* appears to wrap after enough input.

English text escapes the deadlock because code units == cells: the estimate grows as soon as a line exceeds the box width. (A milder variant is still reachable in English with pathological word lengths: three words each slightly over half the box width word-wrap to 3 rows while the estimate says 2, and the cap hides the third row.)

Two smaller observations from instrumenting the component:

- `useEffect([draft?.body])` resets the line-count hint to the hard newline count on every keystroke, wiping the `virtualLineCount`-based correction from `onContentChange` one render later.
- The same code-unit estimate feeds `measureAgentInlineNoteHeight`, so planned row heights drift from mounted heights for CJK drafts (the row-windowed stream relies on these agreeing).

### Steps to reproduce

1. `hunk diff` on any change (split layout).
2. Press `c` to start a review note.
3. Type a long CJK sentence without spaces (e.g. repeat 测试 ~40 times).
4. The composer stays one row high and shows only the cursor's visual row (figures 1-2); an English sentence of similar display length wraps instead. Typing past ~2× the visual width suddenly self-heals.

### Expected behavior

The draft composer wraps by terminal cells and grows a row whenever the text exceeds the box width — for CJK exactly as it already does for English, and consistent with how the saved note renders (cell-aware
`wrapText` since #567). The height estimate, the editor's actual wrapping, and `measureAgentInlineNoteHeight` should all agree.

A natural fix is to give `draftVisualLineCount` the same cell-aware word wrap semantics as `wrapText` (or share one helper), so the estimate always matches the editor's real wrapping and the height-1 deadlock never forms.

### Version

0.18.0-beta.0

IAMLEIzZ:
Fixes #681

## Summary

The draft review-note composer estimates its height with `String#length` (UTF-16 code units) in `draftVisualLineCount`, while the textarea wraps by terminal cells. For CJK text (1 code unit = 2 cells) the estimate stays at one row — and since the editor clamps its wrap count to the viewport height, a one-row-high composer never starts wrapping at all: the text scrolls horizontally and everything before the cursor disappears (see the issue for screenshots). English escapes this because code units == cells, so the estimate grows in step and un-clamps the editor.

- give the composer `wrapMode="char"` so the wrapped row count is computable exactly (the default word wrap follows UAX #14 — e.g. kinsoku punctuation rules — which a small estimator cannot mirror)
- rewrite `draftVisualLineCount` to pack grapheme clusters into terminal cells (a tab counts as 2 cells, matching the editor), with a fast path for printable ASCII lines
- drop the stale line-count hint state: the estimate is now exact, and the correction channel was already defeated by the viewport clamp and a resetting effect
- resize the composer in the same frame as each edit (`flushSync` around `onInput`), so the editor never sits in a content-taller-than-viewport state

Planned heights (`measureAgentInlineNoteHeight`) and mounted heights now come from the same exact function, keeping the row-windowed stream in lockstep.

One behavior change to call out: in the composer, English text now wraps at the exact box width (mid-word if needed) instead of at word boundaries.Saved note cards are unchanged — they still word-wrap via `wrapText`.

## Repro

With a ~76-cell composer box and a 60-char / 120-cell CJK draft:

|        | Row estimate        | Result                              |
| ------ | ------------------- | ----------------------------------- |
| Before | `ceil(60/76)` = 1   | viewport clamps wrapping, text scrolls horizontally, only the tail visible |
| After  | 2 (cluster packing) | full text visible, wrapped          |

## Testing

- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- `bun test ./src ./packages ./scripts ./test/cli ./test/session` — 1970 pass; the one failure (`src/ui/AppHost.watch.test.tsx`) also fails on a clean checkout of the base commit, unrelated to this change
- `bun test test/pty/notes.test.ts` — 15/15, including a new real-PTY case: type 86 cells of CJK, assert head and tail stay visible, save with Ctrl-S

New coverage:

- `AgentInlineNote.test.tsx`: table-driven estimator cases (CJK, emoji, ZWJ sequences, combining marks, tabs, hard newlines, odd widths), plus parity tests that mount a real `TextareaRenderable` and assert its `virtualLineCount` equals the estimator for 23 inputs × 4 widths
- AppHost stream-level regression: press `c`, type long CJK, assert the head of the text stays visible
- PTY regression as described above

### DIFF
--- a/.changeset/draft-note-cell-wrap.md
+++ b/.changeset/draft-note-cell-wrap.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Wrap draft review notes by terminal cells instead of scrolling horizontally, so long CJK notes stay fully visible while typing; previously the composer stayed one row high and hid everything before the cursor.
--- a/src/ui/AppHost.interactions.test.tsx
+++ b/src/ui/AppHost.interactions.test.tsx
@@ -2822,6 +2822,72 @@ describe("App interactions", () => {
     }
   });
 
+  test("draft note wraps long CJK input instead of scrolling it out of view", async () => {
+    const setup = await testRender(<AppHost bootstrap={createBootstrap()} />, {
+      width: 160,
+      height: 40,
+    });
+
+    try {
+      await flush(setup);
+
+      await act(async () => {
+        await setup.mockInput.typeText("c");
+      });
+      await flush(setup);
+
+      const body =
+        "这个包主要是为了在普通的chatmodel外面包一层,在外层把toolcallid统一转换,方便后续处理";
+      for (const chunk of body.match(/.{1,12}/g) ?? []) {
+        await act(async () => {
+          await setup.mockInput.typeText(chunk);
+        });
+        await flush(setup);
+      }
+
+      const frame = setup.captureCharFrame();
+      expect(frame).toContain("Draft note");
+      expect(frame).toContain(body.slice(0, 10));
+      expect(frame).toContain(body.slice(-4));
+    } finally {
+      await act(async () => {
+        setup.renderer.destroy();
+      });
+    }
+  });
+
+  test("draft note survives a large burst of input in one chunk", async () => {
+    const setup = await testRender(<AppHost bootstrap={createBootstrap()} />, {
+      width: 160,
+      height: 40,
+    });
+
+    try {
+      await flush(setup);
+
+      await act(async () => {
+        await setup.mockInput.typeText("c");
+      });
+      await flush(setup);
+
+      // One synchronous burst, the shape chunked pastes and key repeats take.
+      const text = "the quick brown fox jumps over the lazy dog 0123456789".repeat(3);
+      await act(async () => {
+        await setup.mockInput.typeText(text);
+      });
+      await flush(setup);
+
+      const frame = setup.captureCharFrame();
+      expect(frame).toContain("Draft note");
+      expect(frame).toContain(text.slice(0, 10));
+      expect(frame).toContain(text.slice(-6));
+    } finally {
+      await act(async () => {
+        setup.renderer.destroy();
+      });
+    }
+  });
+
   test("draft note saves Ctrl-S when tmux sends CSI-u input", async () => {
     const setup = await testRender(<AppHost bootstrap={createBootstrap()} />, {
       width: 240,
--- a/src/ui/components/panes/AgentInlineNote.test.tsx
+++ b/src/ui/components/panes/AgentInlineNote.test.tsx
@@ -1,0 +1,313 @@
+import { describe, expect, test } from "bun:test";
+import { TextareaRenderable } from "@opentui/core";
+import { createTestRenderer } from "@opentui/core/testing";
+import { testRender } from "@opentui/react/test-utils";
+import { act, useState } from "react";
+import { resolveTheme } from "../../themes";
+import {
+  AgentInlineNote,
+  draftVisualLineCount,
+  measureAgentInlineNoteHeight,
+} from "./AgentInlineNote";
+
+const theme = resolveTheme("github-dark-default", null);
+
+describe("draftVisualLineCount", () => {
+  const cases: Array<[string, string, number, number]> = [
+    // [label, text, width, expected rows]
+    ["empty text", "", 10, 1],
+    ["short ASCII fits", "hello", 10, 1],
+    ["ASCII exact fit", "aaaaaaaaaa", 10, 1],
+    ["ASCII one cell over", "aaaaaaaaaaa", 10, 2],
+    ["long unbroken ASCII", "a".repeat(50), 10, 5],
+    ["word slack packs by cells", "aaaaaa aaaaaa aaaaaa", 10, 2],
+    ["trailing space counts", "aaaaaaaaaa ", 10, 2],
+    ["spaces only", "   ", 10, 1],
+    ["CJK exact fit", "阿斯蒂芬加", 10, 1],
+    ["CJK one cluster over", "阿斯蒂芬加快", 10, 2],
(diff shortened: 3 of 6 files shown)


benvinegar (changes requested):
Requesting changes: bulk paste can re-enter `onContentChange` through `flushSync`, hit “Maximum update depth exceeded,” and truncate the draft. Please make the update non-reentrant and add bracketed-paste coverage.

IAMLEIzZ:
> Requesting changes: bulk paste can re-enter `onContentChange` through `flushSync`, hit “Maximum update depth exceeded,” and truncate the draft. Please make the update non-reentrant and add bracketed-paste coverage.
> 
> _This comment was generated by Pi using GPT-5.6 Sol_

Fixed — dropped the per-keystroke `flushSync` so burst input batches into onerender, and added bracketed-paste + burst-input regression tests. (This crash actually predates the CJK fix; the old hint state flushed the same way.)

IAMLEIzZ:
Both red checks look unrelated to this change:

- **Windows compatibility** fails in test cleanup with `EBUSY: resource busy or locked` on `rmSync` of a temp dir (`AppHost.workspace.test.tsx:34`) — the same failure shows up on other PRs (e.g. runs 31244480282 and 31238789978), so it looks like a recurring Windows flake. Could you re-run the failed job?
- **Vercel** just needs a team member to authorize the fork deployment.

Thanks!!!

benvinegar (changes requested):
Requesting changes: `draftVisualLineCount` still does not exactly match the textarea's width semantics for some grapheme clusters. It measures clusters through `measureClusterWidth`/`string-width`, while OpenTUI's native editor uses Bun's width result. For a 24-cell textarea, this input reproduces the undercount:

```ts
"HEAD-" + ("🇺🇸\u0301").repeat(10) + "-TAIL"
```

`draftVisualLineCount` returns 1, but a real `TextareaRenderable` reports 2 rows. Mounting `AgentInlineNote` at that width allocates one row and omits `-TAIL` from the frame. Keycap and heart emoji with an attached combining mark show the same mismatch (`string-width` reports 1 cell for the cluster while Bun/OpenTUI reports 2).

Please measure with the editor's width semantics and add this input to both the editor-parity table and a mounted visibility regression. Ordinary CJK wrapping, burst input, and bracketed paste otherwise worked in my validation, including the full unit, PTY integration, and TTY smoke suites on a merge with fresh `main`.

IAMLEIzZ:
> Requesting changes: `draftVisualLineCount` still does not exactly match the textarea's width semantics for some grapheme clusters. It measures clusters through `measureClusterWidth`/`string-width`, while OpenTUI's native editor uses Bun's width result. For a 24-cell textarea, this input reproduces the undercount:
> 
> ```ts
> "HEAD-" + ("🇺🇸\u0301").repeat(10) + "-TAIL"
> ```
> 
> `draftVisualLineCount` returns 1, but a real `TextareaRenderable` reports 2 rows. Mounting `AgentInlineNote` at that width allocates one row and omits `-TAIL` from the frame. Keycap and heart emoji with an attached combining mark show the same mismatch (`string-width` reports 1 cell for the cluster while Bun/OpenTUI reports 2).
> 
> Please measure with the editor's width semantics and add this input to both the editor-parity table and a mounted visibility regression. Ordinary CJK wrapping, burst input, and bracketed paste otherwise worked in my validation, including the full unit, PTY integration, and TTY smoke suites on a merge with fresh `main`.
> 
> _This comment was generated by Pi using GPT-5.6 Sol_

Confirmed and reproduced — thanks. `draftVisualLineCount` no longer relies on JS width tables at all: it now measures through a shared native EditBuffer/EditorView, so the count matches the editor by construction on every runtime. Your input is added to both the editor-parity table and a mounted visibility regression; full unit, PTY, and TTY smoke suites pass.

IAMLEIzZ:
Both rounds of feedback are addressed and pushed. Whenever you have a moment, could you take another look? Happy to adjust anything else.Thanks!

benvinegar (approved):
Verified the requested paste-safety and native editor-width fixes, including targeted regressions. Approving.

*This comment was generated by Pi using gpt-5.6-terra*

PR #695, opened 2026-08-09
Fix malformed @@ header in formatHunkHeader fallback

YuriNachos:
## Summary
The fallback branch of `formatHunkHeader` now emits a well-formed unified-diff `@@` header for every Hunk, including the edge case that previously produced a malformed one.

## Root cause
The fallback used `*Lines` (changed-lines count) instead of `*Count` (total lines incl context) for the `@@ -start,count +start,count @@` fields, so a Hunk with context lines got a malformed header.

## Changes
- `src/core/hunkHeader.ts` — use `*Count` in the fallback `@@` construction (matches the field semantics documented in `@pierre/diffs` + the existing primary branch).
- `src/core/hunkHeader.test.ts` (new, bun:test) — proves red→green: a fallback Hunk with `deletionStart:10, deletionCount:4` now yields `@@ -10,4 +10,4 @@` (was `@@ -10,1 +10,1 @@`), plus a pure-addition `@@ -0,0 +1,3 @@` case and a context-suffix case.

## Test plan
- [x] `bun test src/core/hunkHeader.test.ts` — 3 pass, 0 fail.

This change was developed with AI assistance (Claude Code); every changed line was reviewed and understood.

PR #689, opened 2026-08-08
fix(session): report daemon reachability in session list JSON

YuriNachos:
Fixes #527

## Summary
`hunk session list --json` now includes a `daemon` object alongside `sessions` so callers can distinguish an unreachable daemon from a healthy daemon with zero sessions — both previously printed the byte-identical `{ "sessions": [] }` with exit 0.

- Daemon unreachable: `{ "sessions": [], "daemon": { "available": false } }`
- Daemon reachable: `{ "sessions": [...], "daemon": { "available": true } }`

Text-mode `hunk session list` output and the exit code are unchanged in both states, and existing JSON consumers that read `.sessions` keep working (purely additive, non-breaking).

## Root cause
In `runSessionCommand`, the daemon-down `list` branch returned `renderOutput(input.output, { sessions: [] }, …)` and the daemon-up `list` branch returned `renderOutput(input.output, { sessions }, …)`. `renderOutput` serializes the value object as the JSON body, so both produced `{ "sessions": … }` — the reachability boolean from `resolveDaemonAvailability` was discarded before it reached the output, collapsing "daemon down" and "daemon up, zero sessions" into the same bytes.

## Changes
- `src/session/agent/commands.ts` — pass `daemon: { available }` inside the `value` to `renderOutput` for the `list` action in both branches (`available: false` on the down path, `available: true` on the up path). No other action is affected; the return type stays `string`; `src/main.tsx`, exit codes, and CLI flags are untouched.
- `src/session/agent/commands.daemon.test.ts` — update the real-probe strict `toEqual` to the new shape `{ sessions: [], daemon: { available: false } }`.
- `src/session/agent/commands.test.ts` — add two hook-based regression tests (daemon-unavailable and daemon-available JSON shapes).
- `.changeset/session-list-daemon-status.md` — `hunkdiff` / `patch` changeset.

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run format:check`
- [x] `bun run test:theme-contrast`
- [x] `bun run test` (1940 pass / 0 fail, incl. `src/session/agent/commands.test.ts` and `commands.daemon.test.ts`)
- [x] `bun run test:tty-smoke`
- [x] `bun run test:integration` (91 pass / 6 fail — the 6 failures are pre-existing TUI/extension tests that reproduce identically on the clean base `c4739ef` and are caused by the local worktree environment, not this change; the full unit + session suite is green)
[...]

PR #696, opened 2026-08-09
fix(hunkHeader): use per-side total count in the fallback @@ header

YuriNachos:
## Summary
`formatHunkHeader`'s fallback branch emitted a malformed `@@ -l,s +l,s @@` header for any hunk carrying context lines, because it used the changed-lines count (`*Lines`) instead of the per-side total count (`*Count`).

## Root cause
The fallback built the header from `hunk.deletionLines` /`hunk.additionLines`, which count only the changed `+`/`-` lines. The unified-diff `@@` count is the per-side total (context + changed), so any hunk with context was undercounted and malformed.

## Fix
Use `hunk.deletionCount` /`hunk.additionCount` (the per-side totals). Hunks with no context are unchanged.

## Tests
New `hunkHeader.test.ts` covers the context-carrying case (was malformed, now correct) and the no-context case (unchanged).

### DIFF
--- a/src/core/hunkHeader.test.ts
+++ b/src/core/hunkHeader.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, test } from "bun:test";
+import type { Hunk } from "@pierre/diffs";
+import { formatHunkHeader } from "./hunkHeader";
+
+/**
+ * Build a synthesized hunk that has no parsed `hunkSpecs`, so `formatHunkHeader`
+ * falls back to building the `@@ -l,s +l,s @@` header from the numeric fields.
+ *
+ * A changeset transform can legally produce such a hunk: validation only
+ * requires `hunkContent` (`summarizeHunk`), and a transform that supplies the
+ * numeric header fields but no raw specs drives this exact fallback path both
+ * in the review stream (`src/ui/diff/pierre.ts`) and in `summarizeHunk`.
+ */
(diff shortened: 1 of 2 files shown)


YuriNachos:
Heads-up on the red check, so it doesn't read as a problem with this branch: **the failing `Vercel` job is the fork-deployment authorization gate, not a build or test failure.** Its status text is:

> Authorization required to deploy.

and the target URL is `vercel.com/git/authorize?team=Modem&…`, i.e. a request for someone on the Modem team to authorize a deployment from a fork. It is red on **every** open fork PR in this repo right now, mine and other people's alike — sampled just now: #696, #695, #689 (mine) and #686, #682, #678, #676 (not mine), all with the identical message.

Every check a fork PR can actually influence is green here: `Socket Security: Project Report` and `Socket Security: Pull Request Alerts` both pass.

Nothing on this branch can turn Vercel green — it needs the authorize link clicked on your side. Happy to rebase or re-trigger if that helps.

YuriNachos:
Closing this as a duplicate of #695, which was just merged (`63babe0`) with the same fix: `formatHunkHeader`'s fallback now uses the per-side totals (`additionCount`/`deletionCount`) instead of the changed-lines count (`*Lines`), plus the same synthesized-hunk regression test. Sorry for the double PR — #695 and #696 were opened in parallel and addressed the identical bug. No maintainer time needed here.

PR #713, opened 2026-08-11
feat(ui): respect current line when opening in $EDITOR

JammingBen:
Make opening a file in `$EDITOR` respect the line number of the current cursor instead of opening at the start of the selected hunk.

Also move the `AppHost edit-selected-file shortcut` tests into a dedicated file, since I figured they don't really belong in `src/ui/AppHost.sidebar-resize.test.tsx`.

### DIFF
--- a/src/ui/lib/openInEditor.ts
+++ b/src/ui/lib/openInEditor.ts
@@ -2,17 +2,74 @@ import { existsSync } from "node:fs";
 import { basename, resolve, win32 } from "node:path";
 import type { CliRenderer } from "@opentui/core";
 import type { DiffFile } from "../../core/types";
+import type { LineCursor } from "./lineCursors";
 
 export interface EditorCommand {
   command: string;
   args: string[];
 }
 
+type DiffHunk = DiffFile["metadata"]["hunks"][number];
+
+/** The review stream's current line, minus the geometry fields this module never reads. */
+export type EditorLineCursor = Pick<LineCursor, "fileId" | "hunkIndex" | "target">;
+
+/**
+ * Translate an old-side line to the line it maps to in the file on disk.
+ *
+ * Deleted lines have no on-disk counterpart, so they resolve to the position their
+ * replacement occupies, which is where the reader expects the editor to land.
+ */
+function deletionLineToFileLine(hunk: DiffHunk, deletionLine: number) {
+  let deletionCursor = hunk.deletionStart;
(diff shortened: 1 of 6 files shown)


src/ui/lib/openInEditor.ts
```diff
@@ -2,17 +2,74 @@ import { existsSync } from "node:fs";
...
+        return additionCursor + (deletionLine - deletionCursor);
+      }
+
+      deletionCursor += content.lines;
+      additionCursor += content.lines;
+      continue;
+    }
+
+    if (deletionLine < deletionCursor + content.deletions) {
+      return additionCursor;
```
benvinegar: When the cursor is on the second deleted line of a multi-line replacement, this currently opens the first added line. Could we preserve the line's offset within the replacement block?

src/ui/lib/openInEditor.ts
```diff
@@ -2,17 +2,74 @@ import { existsSync } from "node:fs";
...
+/**
+ * Translate an old-side line to the line it maps to in the file on disk.
+ *
+ * Deleted lines have no on-disk counterpart, so they resolve to the position their
+ * replacement occupies, which is where the reader expects the editor to land.
+ */
+function deletionLineToFileLine(hunk: DiffHunk, deletionLine: number) {
+  let deletionCursor = hunk.deletionStart;
+  // A zero-count side names the line before the change, so step past it to land inside the file.
+  let additionCursor = hunk.additionCount === 0 ? hunk.additionStart + 1 : hunk.additionStart;
```
benvinegar, line 26: For a deletion at EOF, this can request a line past the end of the resulting file. Should we clamp it to the final line?

benvinegar (commented):
The main behavior looks good. This just needs a rebase onto `main`; I left two old-side mapping edge cases inline.

src/ui/lib/openInEditor.ts
JammingBen: It seems to be handled gracefully by every editor that supports this feature, and all other editors don't get a line reported anyways.

I might be wrong on that, but I think we would need the true file count, which we can only reliably get with `isPartial: false`, or when reading the file content (e.g. via `readFileSync`). So given the target editors already handle it, I'd rather not change it.

But if you still want me to tackle it, I can do so.

benvinegar:
@JammingBen We did it

PR #714, opened 2026-08-11
feat: support mise as a first-class install path

elucid:
Hunk now ships as a default tool in [Omarchy](https://omarchy.org), installed with `omarchy-mise-install aqua:modem-dev/hunk hunk`. That makes [mise](https://mise.jdx.dev) a first-class distribution path for us, and this PR makes Hunk aware of it.

Worth noting: `mise use -g hunk` already worked before this PR. Community contributors added Hunk to both the [mise registry](https://github.com/jdx/mise/blob/main/registry/hunk.toml) and [aqua-registry](https://github.com/aquaproj/aqua-registry/blob/main/pkgs/modem-dev/hunk/registry.yaml) without our involvement. "supporting mise" just means noticing when we break those entries, and publishing descriptive and verification fields.

## What's here

**1. Suppress the startup update notice for mise installs** (`src/core/updateNotice.ts`)

The omarchy wrapper runs `mise use -g` before exec'ing the binary, so a mise session is already on the newest release by the time Hunk starts. Those installs previously fell through to the `npm` install source and were told to run `npm i -g hunkdiff` — wrong for how they installed, and redundant with an upgrade that had just happened.

Detection uses adjacent `mise` + `installs` path segments. mise's env vars (`MISE_SHELL`, `__MISE_DIFF`) were rejected: they're set by `mise activate` shell integration, not by `mise x`, which is exactly how the omarchy wrapper launches us. Suppression is a named `SELF_UPDATING_INSTALL_SOURCES` policy checked before the dist-tags fetch, so mise installs skip the network entirely.

**2. Attest release archives** (`.github/workflows/release-prebuilt-npm.yml`)

mise verifies aqua-installed tools by default — `aqua.github_attestations`, `aqua.cosign`, and `aqua.minisign` all default to `true` — but we publish no verification material at all, so every mise/aqua install of Hunk is currently unverified.

`actions/attest-build-provenance` now runs before upload. The subject glob mirrors the upload glob rather than matching `*.tar.gz`, so the invariant is "nothing leaves that directory unattested" even if the asset set changes later.

**3. Document the mise path** (`README.md`, `AGENTS.md`, website)

Also scopes the "Node.js 18+" requirement to the npm install, which never applied to the Homebrew, Nix, or mise paths.

The release-verification command is `MISE_AQUA_BAKED_REGISTRY=false mise latest hunk`. The flag matters: mise bakes an aqua registry snapshot into each of its own releases and caches registry sources for a week, so a naive check can pass or fail for reasons unrelated to our release.

## Verification

- `bun test src/core` — 547 pass, 8 skip. The three failures are pre-existing and environment-only: they assert npm-flavored instructions without stubbing `resolveExecutablePath`, so they read `process.execPath` and resolve `nix` on a Nix-store Bun. Verified identical at the base commit, and green in CI, where Bun sits at `~/.bun/bin/bun`
- mise install exercised end to end in a sandbox: `mise use -g hunk` fetched `hunkdiff-darwin-arm64.tar.gz` and ran `hunk --version` → `0.18.0`, no Node involved
- `MISE_AQUA_BAKED_REGISTRY=false` confirmed to actually bypass the baked snapshot
- Confirmed the mise early-return sits *after* `resolveStartupSkillRefreshNotice`, so mise users still receive the one-time skill-refresh notice
- Action pin verified against the GitHub API: `v4.2.2` → `4d101475d8b20a2381f78447822ac1eab6504dd8` (lightweight tag, ref SHA is the commit SHA)

The attestation path itself can only be proven by a real tagged release — the job is gated `if: github.event_name == 'push'`, so `workflow_dispatch` never exercises it.

## Follow-ups, deliberately not in this PR

**Windows support in aqua-registry.** Investigated and proven on a real Windows 11 x64 box with `aqua` directly (aqua-registry closes reports that only reproduce through mise). Result: **GO**, and no `.exe` templating is needed — aqua's `completeWindowsExtToFileSrc` appends `.exe` when `files[].src` has no extension, so our existing `src` already resolves to `hunkdiff-windows-x64/hunk.exe`. Only `supported_envs` is missing.

The entry does need a version split, not a one-line change: Windows assets only start at v0.12.0, and the naive diff fails on v0.11.1 with `the asset isn't found: hunkdiff-windows-x64.tar.gz`. Today the current entry silently installs *nothing* on Windows and exits 0, which is the mise fallthrough. Bundled `skills/` were confirmed to survive extraction.

**The aqua-registry verification block.** Blocked on item 2 shipping — the `signer_workflow` value should be read off a live attestation via `gh attestation verify --format json`, not hand-written.


elucid:
Windows support in `aqua-registry` is now complete

PR #720, opened 2026-08-12
feat(review): navigation intents, command catalog, and the address grammar (Phase 1 PR 3)

benvinegar:
Phase 1 PR 3 — the final Phase 1 PR of the browser-review rebuild (`docs/browser-review-rebuild.md`). Stacked on PR 2 (#719).

## What

Three commits, each green standalone:

1. **One shared navigation walk** — `src/core/review/navigation.ts` plans `selection/move` (`hunk | file | annotated-hunk | annotated-file` scopes, explicit per-scope wrap policy) and `selection/select-file`; both the terminal keyboard path *and* the agent-session comment navigation now plan through it. `src/ui/lib/hunks.ts` deleted and **tombstoned** in the boundary suite.
2. **Command identity split out of the terminal table** — `src/core/commandCatalog.ts` holds id/title/category/default chords/resolution locus as data; semantic commands declare their review effect as *data* with one `lowerAppCommandToReviewIntent`. The terminal's dispatch table is built from the catalog; menus and help read through it. Keyboard behavior byte-identical.
3. **Semantic address grammar** — `src/core/review/address.ts`, `formatReviewAddress`/`parseReviewAddress` over semantic keys (never array indices), round-trip + strict-rejection tests. No consumers yet by design (browser Phase 5, URL fragments Phase 6).

**Repays** (four-part definition, recorded inline in the audit doc): B1, B2, B4, B5, B9, B11 fully for existing sites; B3, B6, B7 core side (browser halves Phase 5); B8 primitive landed with the terminal site deliberately open (see residuals); F1–F3 catalog side; G3 grammar. B10/B12 untouched (Phase 3, wire).

### Navigation: three walks → one

```mermaid
flowchart TB
    subgraph before["Before — same question, three answers"]
        K1["terminal keyboard<br/>ui/lib/hunks.ts cursors + carry rule"]
        S1["agent session navigation<br/>separate walk, parity by comment only"]
        W1["browser (prototype)<br/>would have been the third"]
    end
    subgraph after["After — one planner, callers supply facts"]
        NAV["core/review/navigation.ts<br/>selection/move · selection/select-file<br/>REVIEW_SELECTION_WRAP_POLICY"]
        K2["terminal keyboard"] -->|"ReviewIntent"| NAV
        S2["hunk session comment nav"] -->|"same intent"| NAV
        B2x["browser — Phase 5"] -.->|"same intent, over the wire"| NAV
    end
```

The wrap/clamp asymmetry that was previously *incidental* is now a named policy: plain hunk/file navigation clamps at edges, annotated-file navigation wraps — `REVIEW_SELECTION_WRAP_POLICY` states it once, and the multi-step carry rule for annotated jumps is in the planner with a conformance fixture pinning it.

### Commands: one fused table → catalog + per-client bindings

```mermaid
flowchart LR
    subgraph before2["Before — appCommands.ts fused three things"]
        FUSED["id + title + chords<br/>+ KeyEvent matchers<br/>+ App-state closures"]
        MENUS1["menus/help read the table"]
    end
    subgraph after2["After"]
        CAT["core/commandCatalog.ts<br/>id · title · category · chords · locus<br/>semantic entries: review effect as data"]
        TERM["ui/lib/appCommands.ts<br/>matchers + handlers only,<br/>built FROM the catalog"]
        LOWER["lowerAppCommandToReviewIntent<br/>(one statement of 'which way does ] move')"]
        MENUS2["menus · help · future palette"]
        CAT --> TERM
        CAT --> LOWER
        CAT --> MENUS2
        WEB["browser — Phase 5"] -.->|"renders palette + bindings<br/>from the same catalog"| CAT
    end
```

Catalog↔handler parity is a **compile error**, not just a test: `AppCommandId` is derived from the catalog literal and the terminal handler map is keyed by it. A parity test additionally pins that menus and the dispatch table contain exactly the catalog's built-ins.

### Where each command resolves

```mermaid
flowchart TB
    KEY["key press / menu click"] --> MATCH["terminal matcher (unchanged chords)"]
    MATCH --> LOCUS{"catalog locus"}
    LOCUS -->|semantic| INT["lower to ReviewIntent → planner → store<br/>(navigation, filter, toggle notes)"]
    LOCUS -->|client-local| LOCAL["terminal handler<br/>(scroll, layout, theme, help)"]
    LOCUS -->|host-only| HOST["host effect<br/>(quit, refresh, $EDITOR)"]
    INT --> ALL["every attached client re-renders"]
```

## Residuals recorded in the audit (deliberate)

- **B4 correction**: the audit assumed the terminal rendered "no file" for a filtered-out selection — in fact it kept rendering the selected file. Terminal behavior won; `selectNormalizedSelection` yields `fileKey: null` only when nothing is visible.
- **B8**: `selectNotesByHunk` (owner-index grouping) landed, but DiffPane still groups by range overlap — converting is user-visible and belongs in a behavior-changing PR, not this one.
- **B5**: committed-vs-live filter timing still undecided (planning reads immediate, rendering reads one-render-deferred).
- Two long-standing navigation quirks are now *named* rather than incidental (edge-hunk re-reveal; annotated-file ring entry at index 0).
- `expansion/toggle` and the note-draft intent are the two named gaps in the catalog's semantic set (`SEMANTIC_COMMANDS_WITHOUT_REVIEW_EFFECT`) — Phase 2 fills them, which is what lets `z` and `c` reach a remote client.

## Design notes (fresh implementation, not a port)

- Navigation plans over an explicit `ReviewNavigationModel` + caller-supplied `ReviewAnnotationIndex` — core never reaches into a renderer's diff model for facts it can't own.
- Catalog placement: `src/core/commandCatalog.ts` — UI vocabulary doesn't belong in `core/review` (pure review semantics), and `src/ui` is unreachable from a browser bundle. No react/opentui/node imports; string chords only. **Phase 5 note**: the web boundary gate's allowed-targets list must add this path when the browser adopts it.
- Address grammar: slash path with percent-encoded identifiers, fragment-safe, strict parsing.

## Gates

- `bun run typecheck`, `bun run lint` (0/0), `bun run format` — clean; empty changeset.
- `bun test`: 2457 pass / 5 fail — failures identical to the base commit's (dev-dep ×2 + three known flakes, each passing in isolation); strictly fewer failures than baseline's run.
- `bun run test:integration` (PTY): 105/106; the one failure reproduces identically on the base commit. PTY files untouched.
- Boundary suite green with the first tombstone entry (`src/ui/lib/hunks.ts` — B1).
- Real source invocation smoke (`bun run src/main.tsx -- diff`) renders correctly.

---
_Generated by [Claude Code](https://claude.ai/code/session_018L6h5GBz6RAxRXbgUS4mx4)_


benvinegar:
Independent model-based review completed against this PR's full diff. **No correctness bugs survived verification**: the reviewer hand-traced the navigation walk against every conformance fixture plus independently-constructed edge cases (empty visible set, single file, filtered-out selection, zero-hunk files), verified command precedence is byte-identical to the pre-split table, confirmed the address grammar's round-trip/strict-rejection behavior, confirmed every tombstoned symbol is genuinely absent, and confirmed the conformance consumer drives the real production entry point rather than a parallel copy.

Two minor findings, both addressed in 48a962e:

1. **`lowerAppCommandToReviewIntent` has no production caller yet** — the terminal's handlers read the catalog's declared scope/direction but build intents inline, so the lowering and the closures could silently diverge (only half-covered by tests). Recorded as an F2 residual in the audit doc: it closes when the lowering gains its second consumer (the Phase 5 palette / wire command path), and a review-effect parity check is named as the missing test.
2. **Three new header comments used exactly the passive framing AGENTS.md bans** ("The one … rule/matcher", "The single constructor…") — rewritten in the active register, including one pre-existing offender from the previous PR.

CI re-running on the new head; ready to merge once green.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #726, opened 2026-08-13
feat(extensions): let extensions mark character ranges inside diff lines (API v5)

elucid:
## What

Extensions can now mark **character ranges inside diff lines**, painted by Hunk inside its own rendering — syntax highlighting, word diff, and layout stay intact, and the marked characters get a background the host guarantees is visible.

```ts
hunk.registerLineHighlighter({
  id: "todos",
  highlight({ file }) {
    // Scan file.patch (or a readDocument result) and return marks.
    return [{ side: "new", line: 12, range: [4, 8], tone: "warning" }];
  },
});
```

This closes the one gap between "an extension can jump the review to a hunk" and "an extension can show the user *which characters* it is talking about". Until now the only nearby lever was `registerFileView`, which replaces a file's whole presentation to color a substring.

## Design

- **Source coordinates, never rows.** Marks are addressed by `(side, 1-based line, [start, end))` in UTF-16 code units of the raw line text — what `indexOf`/`RegExp.exec` return against `file.patch`. That addressing survives split vs stack, wrapping, horizontal scrolling, and collapsed-context expansion; extensions never learn Hunk's row model. Offsets landing inside an emoji or CJK glyph widen outward to the whole cluster.
- **Tones, not colors.** A color would be the extension's problem to get right, and it cannot be: a background that reads on a context line is invisible on an added line's green — the exact failure that sank an earlier attempt at this feature. The host resolves each tinted tone (`match`, `info`, `warning`, `error`) against the actual background of each marked line until it clears a perceptual-distance floor well above word-diff emphasis, backing off only before the code on top would stop being readable. `current` renders as reverse video — the `less`/vim convention for the active hit. Verified against every built-in theme × line kind by tests.
- **Paint-time by construction.** Highlights split spans and change colors, never text, so row plans, geometry, wrapping, and the highlighted-diff caches never learn they exist — a highlight change is a repaint, never a re-plan.
- **Pull-based invalidation.** `highlight()` is a pure derivation of `(file, epoch)`: bounded by the same timeout/concurrency discipline as file views, cached until `ctx.highlights.refresh(id, { fileId? })` re-derives it. No host-held mark state to go stale across a reload.
- **Contained failure.** Caps reject oversized results whole; invalid entries drop with one attributed warning; a throwing/timed-out highlighter costs that file's marks and nothing else. The failure mode is always "no marks", never a broken review.

## What this unlocks

A search extension is the motivating consumer (kept out of this PR), but it is the weakest item on the list:

- **Diagnostics inline** — `tsc`/`eslint`/`clippy`/`ruff` output mapped onto the exact columns of changed lines: "this PR introduces this error, *here*", during review rather than in CI later.
- **Agent-driven attention.** With the session daemon already brokering agent commands into live sessions, an agent answering "where does this change behavior?" can light up the exact ranges in the user's terminal as it explains — today it can only scroll to hunks or leave comments. Nothing else in the ecosystem can do that.
- **Trojan-source defense** — bidi controls, homoglyphs, and zero-width characters are invisible *by construction* in a diff viewer; marking those ranges is a security feature a review tool is uniquely placed to offer.
- **Secret scanning** — mark the token that tripped the rule, not the line number in a side pane.
- **Coverage** — uncovered added ranges straight from lcov.
- **Provenance** — which ranges an agent wrote versus a human, squarely in Hunk's stated purpose.
- **Repo conventions** — a `.hunk/extensions/` highlighter marking banned APIs or stray `console.log` for every reviewer on the team.

## Notes

- Bumps the extension API version; the authoring guide (`docs/extensions.md`) gains a typechecked section, the website API page mirrors it, and `docs/line-highlights-proposal.md` records the full design review, including why paint-time placement is load-bearing and how each open question was resolved.
- Includes a shared-policy refactor: file-view invalidation epochs and the bounded `readDocument` capability are extracted into shared modules (`scopedEpochs.ts`, `extensionDocumentReader.ts`) so highlights and file views cannot drift apart.
- Interactive sessions only by construction — the static pager fallback never runs extension code.

## Performance

Verified against `origin/main` with interleaved A/B benchmarking (repo harness + targeted measurements; noise floor calibrated on byte-identical code):

- **Zero-highlighter rendering is unregressed** — 0.999× on ~8s of giant-single-file row rendering, 0.982× on the 180-file stream scroll benchmark; with no highlighters registered the hook returns one shared empty-map singleton and rows take a single early-out branch.
- **Nothing entered the row-plan/geometry cache inputs** — all nine planning/measurement modules are byte-identical to main (verified by content hash), so a mark change is a repaint, never a re-plan.
- **Capped worst cases are sane in absolute terms**: index build at the 2,000-marks/file cap is ~10ms once per file per mark change (memoized, never per frame); span painting at the 100-ranges/line cap is ~7µs/row → well under a millisecond per 40-row frame.
- The verification also surfaced the one theoretical hole in the "paint never moves geometry" contract (a column cut inside a wide glyph); it is now enforced with a runtime guard and pinned by exhaustive width-neutrality tests over CJK/ZWJ/combining/tabbed lines.


benvinegar:
Overall, I think line highlighters are a good addition to the extension system. The core design fits Hunk well: extensions use source coordinates and semantic tones, Hunk owns the rendering, and highlights stay out of Pierre's row planning and review geometry. The registry, trust, rollback, and failure-attribution paths also follow the existing extension architecture.

A few things I think we should fix or revisit before merging:

1. **Old highlights can briefly paint new content after a reload.** `useLineHighlights` keeps the previous `fileId -> marks` map while the replacement files are prepared, and `DiffPane` applies those marks to the current file by ID. File IDs usually survive reloads, so stale offsets can land on unrelated characters. File views already synchronously check file and registration identity before using a cached result; highlights need the same protection.

2. **The caps do not bound aggregate work.** The 2,000-per-file and 100-per-line limits apply to each highlighter separately, then all highlighters are merged without another cap. The paint path scans ranges repeatedly, so many individually valid registrations can create quadratic work. I measured roughly 745 ms to paint one row with 10,000 overlapping ranges. Very large arrays containing only invalid entries also avoid the current valid-entry cap.

3. **Preparation has an all-or-nothing publication barrier.** Highlighters run serially within each file, only four files progress at once, and results are not published until the whole changeset finishes. With 100 files and one timing-out highlighter, good marks can take around 37 seconds to appear. Publishing completed files incrementally would make failure containment match the public promise more closely.

4. **The 512-entry cache can break file-scoped refresh.** Above 512 active `(file, highlighter)` pairs, refreshing one file can cause an eviction cascade and rerun unrelated files. Active-generation results probably need separate retention from any bounded historical cache.

5. **A couple of advertised guarantees are stronger than the behavior.** Tinted marks are deliberately skipped on transparent context backgrounds, despite the release note promising contrast on every line kind. Standalone bidi and zero-width ranges also produce no paint because they occupy no terminal columns, so the Trojan-source/zero-width use case is not currently delivered. We should either add a visible fallback or narrow those claims.

Smaller API questions for later: local IDs containing `:` are accepted but become ambiguous with qualified IDs, and cross-extension refresh lets one extension trigger another extension's callback. Both deserve an explicit ownership rule before this surface settles.

The implementation is otherwise thoughtfully structured, and the size looks mostly justified by the tests, docs, Unicode handling, and containment work. I ran typecheck, the pack/API consumer check, the focused tests, and the full PTY integration suite successfully.

elucid:
@benvinegar hit merge before seeing your feedback. will address feedback in follow up PR

Issue #247, opened 2026-05-08
Proposal: Streaming pager mode for `git log -p` and other large patch streams


bezhermoso:
> **How this was put together:** The design and the prototype implementation were developed in collaboration with a coding agent (Claude Code). I drove the direction, made the judgment calls on scope and trade-offs, and reviewed the code; the agent did the codebase archaeology, drafted the implementation, and helped pressure-test the design. Disclosing because it shaped how this got from "the pager feels broken on `git log -p`" to a working stack of branches in roughly a sitting, and that's worth being transparent about.
>
> **Prototype: https://github.com/bezhermoso/hunk/tree/feat/pager-stream-prototype**

## Problem

When Hunk is configured as Git's pager (`GIT_PAGER=hunk`, `core.pager=hunk`, or invoked as `hunk pager`), it works well for the kind of input most pagers see day to day: a single `git diff`, a single `git show`, a `git format-patch` payload. As soon as the input gets large i.e. running `git log -p` against an old monorepo, the experience falls apart in three distinct ways:

1. **Time to first paint is bound to total stream size.** Hunk reads stdin to completion before doing anything else. Nothing renders until the entire stream has been buffered, decoded, and parsed. On a long-lived repository this can mean a multi-second to multi-minute blank screen, even though the user only wants to scan the first few commits.

2. **Peak memory grows with the input, and the legacy path doubles it.** The full pager input lives in memory as a single decoded string, *and* the parsed metadata derived from it lives alongside, so peak memory is roughly twice the input size. Multi-gigabyte streams thrash the runtime or fail outright. There is no upper bound and no graceful degradation.

3. **Commit metadata vanishes.** Hunk's patch model is per-file. `git log -p` is a per-commit format: commit headers (sha, author, date, message body) sit between the file diffs. The current parser has no concept of these boundaries, so commit metadata gets dropped from stdout. Reading a commit log through Hunk shows a flat stream of file diffs with no indication of which commit produced which change.

The combined effect is that Hunk is currently not a viable pager replacement for the workflow most pagers exist to serve in a large repository: skimming `git log -p` for recent activity. The user has to fall back to `less`, which keeps Hunk siloed to the small/single-changeset case.

## Proposal

Treat pager mode as a stream end-to-end and surface commit context inline with the diff the way `git log -p` does in `less`. I propose the following concepts, each with its own concerns, that together replace the single "buffer then parse" pipeline.

### Streaming stdin I/O

[...]

PR #745, opened 2026-08-14
feat(cli): add --color-only filter mode for git diffFilter

### LINKED ISSUE #575: Non-interactive `--color-only` filter mode for `interactive.diffFilter` (`git add -p`)

Add a non-interactive filter mode, e.g. a `--color-only` / `--filter` flag, or a dedicated `hunk filter` subcommand that reads a unified diff from stdin, applies Hunk's syntax highlighting and diff coloring, writes the result to stdout as an ANSI-colored diff, and exits with no interactive UI or TTY dependency.

Intended use:
```
gitconfig
[interactive]
    diffFilter = hunk <the-new-flag-or-subcommand>
```

### Why?

Hunk works well as `core.pager`, but there's one place in the standard Git workflow it can't reach today: g`it add -p` (and `git stash -p`, `git reset -p`, etc.). These interactive-staging commands don't send hunks to the pager, they read them back through `interactive.diffFilter`, which has a fundamentally different contract:

- A `pager` owns the terminal (a TTY), renders a full-screen UI, and waits for user navigation.
- A `diffFilter` is a non-interactive text transform: Git pipes one hunk's diff to the filter's stdin, expects colorized diff text back on stdout, and re-parses that output to drive its own y/n/s/q prompts. stdout is a pipe, not a TTY, and the filter must not read stdin interactively or hold the terminal.

Because every Hunk command is an interactive viewer, there's currently no way to wire Hunk into `interactive.diffFilter`. Doing so causes `git add -p` to hang, since Hunk grabs the terminal and blocks on input Git needs for its own prompts. This is the gap `delta` closes with `delta --color-only`. Users attempting to migrate from delta to hunk must keep delta for their diffFilter.

### How? (optional)

This shares machinery with #247 (streaming pager mode). Both require a non-interactive, stdout-oriented render path rather than the interactive TUI. The bounded patch sniffer and per-file chunker proposed there are exactly the pieces a filter mode would render from. A `--color-only` filter is essentially "the streaming render path, minus the pager UI, one hunk at a time, straight to stdout." If #247 lands its incremental parse/render pipeline, exposing a filter entry point on top of it should be a small addition rather than a parallel implementation.

YuriNachos:
## What

`hunk --color-only` reads a unified diff from stdin, applies Hunk's diff coloring and syntax highlighting, and writes the ANSI-colored diff to stdout — no interactive UI, no TTY, no alternate screen. Also available as `hunk pager --color-only`.

## Why

Closes #575. Hunk works great as `core.pager`, but `git add -p` / `git stash -p` / `git reset -p` reach their diffs through `interactive.diffFilter` — a non-interactive text transform that Git re-parses to drive its own prompts. Pointing that config at any interactive Hunk command hangs the staging flow, so users migrating from delta have to keep delta installed just for its `--color-only` (`git config interactive.diffFilter "delta --color-only"`). @jesseleite also noted in #575 that a non-interactive colored render enables piping themed hunk output into an fzf preview window without a wrapper script.

```gitconfig
[interactive]
    diffFilter = hunk --color-only
```

## How

- New thin serializer, `src/ui/colorOnlyFilter.ts`, built on Hunk's existing parse/highlight stack (`loadAppBootstrap` → Pierre `loadHighlightedDiff`/`buildStackRows` → theme palettes). There is no second diff engine: the raw stdin lines stay the source of truth and the parsed model only picks colors. That matters because `git add -p` re-parses filter stdout — after stripping ANSI escapes, the output is byte-identical to the input (tabs included). Highlight spans are trusted only when they reconstruct the line exactly; anything that fails to align falls back to whole-line theme colors, and non-diff input passes through unchanged so a misconfigured filter can never corrupt a Git pipeline.
- Startup wiring follows the existing static-pager plan (config-layer themes and custom themes resolve; extensions never load), plus a friendly error when stdin is a TTY.
- Complementary to #686: that PR exports the library-level static renderer (`hunkdiff/static`), while this PR ships the CLI flag the issue asks for, built on the same internal rendering path. If #686 lands, `--color-only` can adopt the exported entry without changing its contract.

## Verification

- `bun run typecheck`, `bun run lint`, `bun run format:check` clean; 17 new unit tests across the renderer, CLI parsing, and startup wiring (`bun test src/ui/colorOnlyFilter.test.ts src/app/startup.test.ts src/core/cli.test.ts` → 150 pass / 0 fail), including the strip-ANSI-equals-input invariant, byte-identical passthrough for non-diff input, recoloring of pre-colored (ANSI) diff input, and tab-indented content bytes.
- Byte-fidelity proof: `diff <(hunk --color-only < sample.patch | strip-ansi) sample.patch` → identical, for plain and pre-colored input.
- `git add -p` end-to-end with `interactive.diffFilter = hunk --color-only`: Git re-parses the colored hunk and drives its prompt normally (verified with `printf 'q\n' | git add -p`).

```console
$ printf 'diff --git a/x.ts b/x.ts\n--- a/x.ts\n+++ b/x.ts\n@@ -1 +1 @@\n-const old = "gone";\n+const fresh = "new";\n' | bun run src/main.tsx --color-only | cat -v
^[[38;2;230;237;243mdiff --git a/x.ts b/x.ts^[[0m
^[[38;2;250;142;137m--- a/x.ts^[[0m
^[[38;2;63;185;80m+++ b/x.ts^[[0m
^[[38;2;173;174;177m^[[48;2;39;43;49m@@ -1 +1 @@^[[0m
^[[38;2;248;81;73m^[[48;2;60;30;33m-^[[0m^[[38;2;255;123;114m^[[48;2;60;30;33mconst^[[0m^[[38;2;121;192;255m^[[48;2;60;30;33m ^[[0m^[[38;2;121;192;255m^[[48;2;79;35;37mgone^[[0m^[[38;2;165;214;255m^[[48;2;60;30;33m "^[[0m^[[38;2;165;214;255m^[[48;2;60;30;33m"^[[0m^[[38;2;230;237;243m^[[48;2;60;30;33m;^[[0m
^[[38;2;63;185;80m^[[48;2;23;51;34m+^[[0m^[[38;2;255;123;114m^[[48;2;23;51;34mconst^[[0m^[[38;2;121;192;255m^[[48;2;23;51;34m ^[[0m^[[38;2;121;192;255m^[[48;2;28;68;40mfresh^[[0m^[[38;2;165;214;255m^[[48;2;23;51;34m "^[[0m^[[38;2;165;214;255m^[[48;2;23;51;34m"^[[0m^[[38;2;230;237;243m^[[48;2;23;51;34m;^[[0m
```

Closes #575. Related: #686, #247.

### DIFF
--- a/.changeset/lucky-moles-color.md
+++ b/.changeset/lucky-moles-color.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": minor
+---
+
+Add `hunk --color-only` to color a unified diff piped on stdin without the interactive UI, for `git interactive.diffFilter` (`git add -p`) and fzf-style diff previews.
--- a/src/app/startup.test.ts
+++ b/src/app/startup.test.ts
@@ -287,6 +287,72 @@ describe("startup planning", () => {
     expect(loaded).toBe(false);
   });
 
+  test("routes color-only pager stdin through the filter plan", async () => {
+    let loaded = false;
+    const patchText = "diff --git a/a.ts b/a.ts\n@@ -1 +1 @@\n-old\n+new\n";
+    const customThemes = [{ id: "custom", base: "github-light-default", text: "#123456" }];
+
+    const plan = await prepareStartupPlan(["bun", "hunk", "--color-only"], {
+      parseCliImpl: async () => ({ kind: "pager", colorOnly: true, options: { theme: "custom" } }),
+      readStdinText: async () => patchText,
+      looksLikePatchInputImpl: () => true,
+      stdoutIsTTY: false,
+      env: { TERM: "xterm-256color" },
+      resolveRuntimeCliInputImpl: (input) => input,
+      resolveConfiguredCliInputImpl: (input) =>
+        createTestConfigResolution(
+          {
+            ...input,
+            options: { ...input.options, theme: "custom" },
+          },
+          { customThemes },
+        ),
+      loadAppBootstrapImpl: async () => {
+        loaded = true;
+        throw new Error("unreachable");
+      },
+    });
+
+    expect(plan).toEqual({
+      kind: "color-only",
+      text: patchText,
+      options: { theme: "custom", pager: true },
+      customThemes,
+    });
+    expect(loaded).toBe(false);
+  });
+
+  test("keeps non-diff stdin on the color-only plan so the filter can pass it through", async () => {
+    let loaded = false;
+    const text = "* main\n  feature/demo\n";
+
+    const plan = await prepareStartupPlan(["bun", "hunk", "--color-only"], {
+      parseCliImpl: async () => ({ kind: "pager", colorOnly: true, options: {} }),
+      readStdinText: async () => text,
+      looksLikePatchInputImpl: () => false,
+      stdoutIsTTY: false,
+      env: { TERM: "xterm-256color" },
+      resolveRuntimeCliInputImpl: (input) => input,
+      resolveConfiguredCliInputImpl: (input) => createTestConfigResolution(input),
+      loadAppBootstrapImpl: async () => {
+        loaded = true;
+        throw new Error("unreachable");
+      },
+    });
+
+    expect(plan.kind).toBe("color-only");
+    expect(loaded).toBe(false);
+  });
+
+  test("rejects color-only filter mode when stdin is an interactive terminal", async () => {
+    await expect(
+      prepareStartupPlan(["bun", "hunk", "--color-only"], {
+        parseCliImpl: async () => ({ kind: "pager", colorOnly: true, options: {} }),
+        stdinIsTTY: true,
+      }),
+    ).rejects.toThrow("reads a unified diff from standard input");
+  });
+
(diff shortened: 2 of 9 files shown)


benvinegar:
Thanks for tackling this—the overall approach looks good, and the focused tests pass locally. I found a few issues to address before merging:

- The PR needs rebasing onto current `main`; `startup.ts` conflicts, and three imports were renamed. Please preserve main’s lazy startup loading.
- `process.exit(0)` immediately after `stdout.write()` truncates large piped output. I reproduced `git add -p` failing with `mismatched output from interactive.diffFilter`.
- Stripping multiline OSC/DCS sequences can consume newlines, breaking Git’s required one-to-one line correspondence.
- Reading stdin through `Response.text()` replaces invalid UTF-8 bytes, so the byte-fidelity guarantee doesn’t hold.
- `--color-only` should live in the shared CLI metadata so generated docs include it.
- Concatenated non-Git unified diffs misclassify the second file’s headers.

Happy to recheck after an update.

PR #747, opened 2026-08-15
perf(spike): measure whether a Rust core pays for itself

benvinegar:
Option 1 of the runtime evaluation was to keep the TypeScript surface and
move hunk's hot paths into a native module. That only makes sense if the
speedup survives the FFI boundary, so this measures it rather than assuming
it.

`crates/hunk-core` ports the two paths a native core would own first:
terminal width measurement (`src/ui/lib/text.ts`) and word-level intraline
diff (today folded into Pierre's highlighting pass) on imara-diff's
histogram algorithm. Nothing in `src/` imports it; it exists to produce a
number.

The number is that the computation is 8x faster and the boundary gives most
of it back. On a real 66k-line changeset, width measurement runs at 34 ns/line
natively against the tuned TypeScript's 272, but only reaches 156 ns/line
through bun:ffi -- marshalling is 57% of the batched path, because handing
UTF-16 JS strings to a UTF-8 library costs more than the library's work. A
1.7x batched win does not justify a second toolchain and per-platform
prebuilts. The generalisation is that a native core pays only when it owns
the text rather than borrowing it, which argues for a Rust review core over
native helpers under TypeScript ones.

Parity is exact on 63,636 real diff lines and a 60-case adversarial unicode
corpus. Reaching that required porting string-width's rules rather than
approximating them; four real bugs surfaced and are covered by regression
tests. The residual 1.79% fuzz mismatch has one known cause, recorded in the
module header and README: Extended_Pictographic is approximated by block
ranges instead of generated from the UCD, so unassigned emoji-block code
points paired with a variation selector measure wide.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01SxpS9muCVHNbSdsA1y8zF8


benvinegar:
Closing this completed experiment because its own measurements conclude that the practical FFI gain does not justify a second toolchain and platform prebuilts. The production startup work landed separately in #750, while the remaining Rust crate was intentionally unshipped research.

PR #774, opened 2026-08-16
fix(commands): construct complete synthetic key events

### LINKED ISSUE #767: Construct complete OpenTUI KeyEvent instances for synthetic command events
## Problem

`synthesizeKeyEvent` returns a plain object asserted as OpenTUI's `KeyEvent`.

The synthetic object contains the chord fields used by current matching, but omits required event fields such as `eventType` and `source`. Its `preventDefault` and `stopPropagation` methods also do not update the state exposed by real `KeyEvent` getters.

The current round-trip test only verifies chord matching. Synthetic events are also passed to command matchers and programmatically invoked handlers, which may legitimately inspect the rest of the `KeyEvent` contract.

## Proposed direction

Construct the exported OpenTUI `KeyEvent` class with a complete parsed-key payload instead of asserting a partial object.

Use explicit synthetic defaults such as `eventType: "press"` and `source: "raw"` unless a more appropriate source is available.

## Acceptance criteria

- `synthesizeKeyEvent` returns a real `KeyEvent` instance without a double assertion.
- Existing chord matching behavior remains unchanged.
- `eventType` and `source` are populated.
- `preventDefault` and `stopPropagation` update the corresponding observable state.
- Extend tests beyond chord round-tripping to cover the complete synthetic event behavior.

nightcityblade:
## Problem and impact

Synthetic command events were partial objects asserted as `KeyEvent`. Matchers or programmatic handlers that inspected required event metadata or propagation state could observe missing or inert behavior.

## Approach

Construct OpenTUI's exported `KeyEvent` with the existing chord fields plus explicit `press` / `raw` synthetic defaults. This keeps chord matching unchanged while using OpenTUI's real prevention and propagation state. No command-dispatch or public extension API behavior is otherwise changed.

This belongs in the existing core command-key helper because it owns synthetic events used by conflict detection and command invocation.

## Verification

- `bun test src/lib/commandKeys.test.ts` (3 passed)
- `bunx oxfmt --check src/lib/commandKeys.ts src/lib/commandKeys.test.ts .changeset/complete-synthetic-key-events.md`
- `bunx oxlint src/lib/commandKeys.ts src/lib/commandKeys.test.ts --deny-warnings`
- `bun run typecheck`

Tested on macOS. No visual evidence is applicable because this changes the internal event contract without altering rendering.

Fixes #767

### DIFF
--- a/.changeset/complete-synthetic-key-events.md
+++ b/.changeset/complete-synthetic-key-events.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Provide complete key event data to command matchers and programmatically invoked handlers.
--- a/src/lib/commandKeys.test.ts
+++ b/src/lib/commandKeys.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, test } from "bun:test";
+import { KeyEvent } from "@opentui/core";
 import { matchesKeyChord, parseKeyChord, synthesizeKeyEvent, toKeyChordList } from "./commandKeys";
 
 /**
@@ -22,6 +23,25 @@ describe("synthesizeKeyEvent", () => {
     for (const chord of ["y", "G", "ctrl+shift+m", "f10", "{", "alt+left", ".", "space"]) {
       expect(matchesKeyChord(parsed(chord), synthesizeKeyEvent(parsed(chord)))).toBe(true);
     }
+  });
+
+  test("returns a complete key event", () => {
+    const event = synthesizeKeyEvent(parsed("ctrl+r"));
+
+    expect(event).toBeInstanceOf(KeyEvent);
+    expect({ eventType: event.eventType, source: event.source }).toEqual({
+      eventType: "press",
+      source: "raw",
+    });
+    event.preventDefault();
+    event.stopPropagation();
+    expect({
+      defaultPrevented: event.defaultPrevented,
+      propagationStopped: event.propagationStopped,
+    }).toEqual({
+      defaultPrevented: true,
+      propagationStopped: true,
+    });
   });
 });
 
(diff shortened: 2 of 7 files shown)


benvinegar:
The intended fix here makes sense: constructing a real `KeyEvent` provides the missing metadata and functional propagation state.

However, changing the type-only import to a runtime `@opentui/core` import introduces a headless-path regression. `src/lib/commandKeys.ts` is also imported by `src/extensions/runExtension.ts` for parsing helpers, so compiled static-pager execution now evaluates OpenTUI and extracts its native library even though no synthetic event is created.

This reproduces in `test/cli/compiled-headless-native-lib.test.ts`: current `main` passes all five tests, while this change causes `keeps captured-host static pager rendering OpenTUI-free` to fail with an extracted `.so`.

Could we split the runtime-dependent synthesis into an interactive-only module—for example, `src/ui/lib/syntheticKeyEvent.ts`—and leave `parseKeyChord`/`toKeyChordList` in the headless-safe `commandKeys.ts` module? `appCommands.ts` and `extensionCommands.ts` could import the synthesis helper directly. That preserves the synchronous API and real `KeyEvent` behavior without pulling OpenTUI into headless consumers.

nightcityblade:
Thanks for the precise report. Fixed in `633ba60`: the runtime `KeyEvent` construction now lives in the interactive-only `src/ui/lib/syntheticKeyEvent.ts`, while `src/lib/commandKeys.ts` is back to a type-only OpenTUI import. The interactive command builders import the synthesis helper directly.

Verification:

- `bun run typecheck`
- `bunx oxlint ... --deny-warnings`
- 35 focused command tests passed
- compiled headless suite: 3 passed, 1 Linux-only static-pager case skipped on macOS (CI will exercise it)

PR #776, opened 2026-08-16
fix(session): reject review event frames whose id disagrees with their fields

### LINKED ISSUE #761: Reject review event frames whose ID disagrees with their publication fields
## Problem

Review event parsers validate `eventId`, `generation`, and `stateRevision` independently, but do not verify that the publication address encoded in `eventId` matches the duplicated fields on the frame.

For example, this frame is currently accepted:

```ts
const eventId = reviewEventId("publication", {
  generation: "generation:producer-a:1",
  stateRevision: 1,
});

parseReviewEventFrame({
  eventId,
  generation: "generation:producer-a:2",
  stateRevision: 9,
  payload: {},
});
```

The event ID identifies generation 1, revision 1, while the frame claims generation 2, revision 9.

The same inconsistency can be accepted by the begin, chunk, and end parsers. Chunk assembly compares frames with the begin envelope, but does not establish that the repeated fields agree with the address encoded in the event ID.

## Why this matters

Event IDs are used for SSE resumption and identify the publication position an event describes. Allowing the encoded address and frame fields to disagree means resume identity, publication ordering, and assembled event metadata can describe different positions.

## Proposed direction

Parse the event ID in each event-frame parser and compare its address with the frame's duplicated fields before accepting the frame.

Where possible, construct the returned typed value from the validated fields rather than casting the original record.

## Acceptance criteria

- `parseReviewEventFrame` rejects an event ID whose generation or revision disagrees with the frame.
- `parseReviewEventBegin` rejects the same inconsistencies.
- `parseReviewEventChunk` and `parseReviewEventEnd` reject event IDs whose generation disagrees with the frame.
- Existing valid event frames continue to parse.
- Add adversarial tests covering mismatched generations and revisions.

jacobabahn:
## Summary

Fixes #761.

The review event-frame parsers (`parseReviewEventFrame`, `parseReviewEventBegin`, `parseReviewEventChunk`, `parseReviewEventEnd`) validated `eventId`, `generation`, and `stateRevision` independently but never checked that the publication address encoded in the event id matches the duplicated fields on the frame. A frame whose id names generation 1 / revision 1 while its fields claim generation 2 / revision 9 was accepted, so resume identity, publication ordering, and assembled event metadata could describe different positions.

## Change

Each parser now parses the event id via `parseReviewEventId` and requires its address to agree with the frame's fields:

- `parseReviewEventFrame` / `parseReviewEventBegin` — generation **and** state revision must match.
- `parseReviewEventChunk` / `parseReviewEventEnd` — generation must match (these frames carry no revision).

Returned values are constructed from the parsed address and validated fields rather than casting the raw record (`record as unknown as ...`), so the decoded frame's position agrees with its id by construction.

No production sender is affected: `planReviewEventFrames` always writes agreeing fields, and all in-repo consumers parse server-produced frames.

## Verification

- 5 new adversarial tests covering mismatched generations and revisions (including the exact example from the issue), plus an accept path asserting valid frames still parse.
- `bun test src/session/reviewEventProtocol.test.ts` — 26/26 pass.
- `bun run typecheck` — clean. `bun run lint` — 0 warnings/errors.
- Full canonical suite (`bun test ./src ./packages ./scripts ./test/cli ./test/session`) — 2763 pass; the 2 failures (`DiffPane add-note clicks` UI test, one `verifyPrReleaseNotes` timeout) reproduce on clean `main` in this environment and are unrelated.
- Changeset added (`patch`) per CONTRIBUTING.md.

### DIFF
--- a/.changeset/consistent-review-event-ids.md
+++ b/.changeset/consistent-review-event-ids.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Reject review event frames whose event id names a different publication generation or state revision than the frame's own fields.
--- a/src/session/reviewEventProtocol.test.ts
+++ b/src/session/reviewEventProtocol.test.ts
@@ -20,6 +20,7 @@ import {
   type ReviewEventBeginV1,
   type ReviewEventChunkV1,
   type ReviewEventEndV1,
+  type ReviewEventFrameV1,
 } from "./reviewEventProtocol";
 
 const ADDRESS = { generation: "generation:test:3", stateRevision: 7 };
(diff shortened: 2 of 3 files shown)


jacobabahn:
Closing per request — fix can be reopened or resubmitted later.

PR #786, opened 2026-08-16
fix(ui): isolate OpenTUI scroll reset compatibility

### LINKED ISSUE #771: Stop silently depending on OpenTUI private scroll accumulator state
## Problem

Shift+Wheel handling in `DiffPane` calls OpenTUI's private `ScrollBoxRenderable.resetScrollAccumulators` method through a double assertion.

The reset is needed after Hunk intercepts shifted wheel input so residual fractional scroll state cannot move the vertical review position later. Because the method is private, OpenTUI can rename or remove it without a type error. Optional chaining then silently skips the reset, which can reintroduce the scrolling regression without an obvious failure.

## Proposed direction

Prefer a public OpenTUI API for resetting wheel accumulator state. If upstream does not expose one yet:

- isolate the private access in a named OpenTUI compatibility adapter;
- document the exact runtime invariant and supported OpenTUI version;
- fail clearly when the private method is unavailable rather than silently degrading;
- retain user-visible interaction coverage for shifted vertical and native horizontal wheel events.

The compatibility adapter should be removed once a public upstream operation is available.

## Acceptance criteria

- `DiffPane` no longer performs an inline private-member assertion.
- An OpenTUI upgrade that removes or renames the required operation fails focused tests clearly.
- Shift+Wheel continues scrolling code horizontally without changing the vertical review position.
- The workaround is localized and documents its deletion condition.

atirna:
## Problem

Shift+Wheel has to clear OpenTUI wheel remainders after Hunk reroutes the input horizontally. The previous inline private-member assertion used optional chaining, so an OpenTUI rename could silently let later input move the review viewport.

## Approach

- isolate the OpenTUI 0.5.1 private access in one named compatibility bridge
- document the invariant and removal condition for a future public OpenTUI API
- throw a focused error if the required operation disappears instead of silently degrading
- cover the locked runtime operation and Shift+Wheel handling in both rendered and real-PTY tests

## Validation

- `bun run typecheck`
- `bun run lint`
- `bun run format:check`
- `bun run check:docs`
- `bun run test` (2807 passed, 24 skipped)
- `bun test src/ui/components/panes/DiffPane.test.tsx src/ui/AppHost.interactions.test.tsx --test-name-pattern "resetOpenTuiScrollAccumulators|shift plus (mouse wheel|native horizontal)"` (5 passed)
- `bun test test/pty/layout.test.ts --test-name-pattern "shifted mouse-wheel input"` (1 passed)

Fixes #771


benvinegar:
Thanks, good idea!

PR #795, opened 2026-08-17
Organize src/core into real modules

benvinegar:
## Problem

`src/core` had become a pile of ~30 loose files that any code anywhere could reach into. The worst symptom was `core/types.ts`: 147 files imported it, it mixed four unrelated audiences (the changeset model, CLI inputs, app bootstrap, and re-exported extension contract types), and touching it invalidated a big chunk of the typecheck graph. There was no way to tell which files were "API" and which were internals — everything was public because nothing said otherwise.

```mermaid
flowchart TD
  ui["src/ui"] --> types
  app["src/app"] --> types
  session["src/session"] --> types
  ext["src/extensions"] --> types
  types["core/types.ts — 147 importing files<br/>changeset model + CLI inputs + bootstrap<br/>+ re-exported extension contract types"]
  types -.->|re-exports for| loose["~30 loose files at core/*<br/>config, loaders, terminal, pager, paths, errors, …<br/>all public, nothing says otherwise"]
  loose -.->|some import back| types
```

## What this does

Core is now a set of directories named for what they hold, each with a declared public surface, and `core/types.ts` is gone — every file imports the module that actually declares the thing it wants, so the dependency graph shows real relationships instead of "everything depends on types.ts".

```mermaid
flowchart TD
  ui["src/ui"] --> changeset
  ui --> run
  ui --> process
  app["src/app"] --> bootstrap
  app --> changeset
  app --> run
  session["src/session"] --> changeset
  session --> run
  subgraph core["src/core"]
    bootstrap["bootstrap.ts<br/>AppBootstrap + ReloadContext<br/>28 importers, down from 147"]
    changeset["changeset/<br/>the model + how it loads<br/><i>internal: fromPatch, sidecar,<br/>fileLanguageLookup</i>"]
    run["run/<br/>what a run was asked to be:<br/>inputs, config, command catalog"]
    process["process/<br/>where it executes: terminal,<br/>pager, job control, shutdown"]
    existing["review/ vcs/ theme/<br/>watch/ patch/<br/><i>already coherent, unchanged</i>"]
  end
  bootstrap --> changeset
  bootstrap --> run
```

The module boundaries aren't just convention: a dependency-cruiser rule fails CI if anything outside a module imports its internals, and a reachability rule fails on any `src/` file no entry point can reach — which is how `core/review/address.ts` (zero consumers) got deleted.

## Reviewing this

It's a big diff but a shallow one: everything is `git mv` plus import-path updates. **No exported symbol was renamed and no behavior changed** — the published `hunkdiff/extension` and `hunkdiff/opentui` entry points are untouched. The commits are reviewable independently if the whole thing is too much at once.

Three judgment calls worth an opinion:

- `core/review/annotations.ts` turned out to genuinely use `AgentAnnotation` from `extension-api/types` — the old shell was hiding that edge. The seam test now allows that single file as a containment target rather than pretending the dependency doesn't exist.
- `changeset/loaders.ts` imports the bootstrap type it assembles (`loadAppBootstrap`), which is backwards for the layering. It's a named exception in the rules; moving that function up to `src/app` would retire it, left as a follow-up.
- `AgentCard.tsx` + `agentPopover.ts` appear to be dead (nothing renders them since notes moved into the diff flow) but only their tests keep them alive. They're quarantined in the dead-module allowlist instead of deleted — happy to delete them here if you agree they're gone.

https://claude.ai/code/session_01XcuayyxJrwQd9EHpuxNef1


benvinegar:
@greptile-apps review

PR #808, opened 2026-08-18
feat(review): add go-to-line prompt

### LINKED ISSUE #805: feat(review): add go-to-line prompt

Add a go-to-line flow to the review UI. Pressing `:` opens a small line-number prompt in the status bar (same inline-input pattern as the `/` file filter). A bare number jumps to that line in the selected file by the **current file's** numbering — the same `R` side the note labels use — while an `l` prefix (e.g. `:l42`) targets the **source file's** numbering (the `L` side). Enter jumps the current-line cursor there so `c` can annotate directly, Escape cancels, and a line outside the diff reports a notice naming the side instead of jumping.

### Why?

#662 added a current-line cursor and anchored notes to it, but moving that cursor is still line-by-line (`j`/`k`) or by mouse. In a large changeset, annotating a specific line you already know — e.g. from a stack trace, a CI log, or a colleague's comment — means dozens of `j` presses or reaching for the mouse. A vim-style `:42` makes keyboard-driven annotation precise in one motion, which was the original motivation of #436.

### How? (optional)

- Register `hunk.review.gotoLine` (default key `:`) in the shared command catalog, remappable via `[keybindings]` like every other command.
- The prompt reuses the status bar's inline prompt input; digits plus an optional leading side letter.
- On submit, resolve the typed line through the existing shared `revealLine(fileId, side, n)` path — the same one the session `navigate` command uses — so scrolling, selection, and the line cursor all land consistently. A line with no rendered row degrades to its hunk or reports "not part of the diff"; no new navigation machinery is introduced.
- I have this implemented with AppHost interaction tests and a PTY test (jump to line 62 across hunks, then anchor a note at R62), plus the one integration fallout handled: the bundled vim-navigation example's `:` command-line moves to `;`, since built-ins shadow extension chords by design.

IAMLEIzZ:
Closes #805

## Summary

Adds a go-to-line flow to the review UI: `:` opens a small line-number prompt in the status bar (the same inline-input pattern as the `/` file filter), and Enter jumps the current-line cursor to that line in the selected file so `c` can annotate it directly. A bare number uses the current file's numbering (the `R` side the note labels use); an `l` prefix such as `:l42` targets the source file's numbering (the `L` side). Escape cancels, and a line that is not part of the diff reports a transient notice naming the side instead of jumping silently.

The jump resolves through the existing shared `revealLine` path that the session `navigate` command uses, so scrolling, selection, and note targeting all land consistently. Line-level navigation is also decoupled from the `cursor_line` display option: when the marker is off, the target cursor is synthesized on demand from the render plan (no full stop-list rebuild), so jumping and annotating stay precise instead of degrading to the hunk.

## Notes

- `hunk.review.gotoLine` is registered in the shared command catalog with `:` as the default chord, is remappable via `[keybindings]`, and closes an open menu when triggered.
- The bundled vim-navigation example moves its command line from `:` to `;`, since built-in chords shadow extension chords by design.
- Failure paths are explicit: lines outside the diff, hunk-only degraded targets, and an empty file selection each report a transient notice, and notices now outrank the `filter=…` summary so they stay visible while a file filter is active.

## Testing

- AppHost interaction coverage: prompt opening without key leakage, same-chunk `:2` delivery, non-digit stripping, Escape cancel, new/old side targeting, closing an open menu, `cursor_line = "off"` jump and annotate precision, selected-file targeting, and the two-stage filter Escape.
- One real-PTY end-to-end test: jump to line 62 across hunks and anchor a note at R62.
- `useTerminalReview` contract tests updated for the synthesized-cursor reveal; vim example unit tests and PTY updated for the move to `;`.
- Full unit suite, typecheck, lint, and format checks pass locally.

### DIFF
--- a/src/ui/hooks/useTerminalReview.ts
+++ b/src/ui/hooks/useTerminalReview.ts
@@ -604,6 +604,20 @@ export function useTerminalReview({
   );
 
   const reconcileLineCursor = useCallback(() => {
+    // A hidden line-level navigation (`cursor_line = "off"`) keeps no stop list, so there
+    // is nothing to resolve against: keep a cursor a direct reveal placed on the selected
+    // hunk, and drop it once the selection moves somewhere else.
+    if (lineCursors.length === 0) {
+      const current = lineCursorRef.current;
+      if (
+        current !== null &&
+        (current.fileId !== selectedFileId || current.hunkIndex !== selectedHunkIndex)
+      ) {
+        applyLineCursor(null);
+      }
+      return;
+    }
+
     // Expansion remeasures before its source text loads, so a toggle records what it wants and
     // this waits for the list that actually carries the revealed rows. Each request survives until
     // it resolves or the next toggle replaces it.
@@ -711,9 +725,10 @@ export function useTerminalReview({
    */
   const revealLine = useCallback(
     (fileId: string, side: "old" | "new", line: number): RevealedLineResult => {
-      const cursor = findLineCursorAt(lineCursorsForRevealRef.current, fileId, side, line);
-      if (cursor) {
-        revealLineCursor(cursor, "reveal");
+      const cursors = lineCursorsForRevealRef.current;
+      const stopped = findLineCursorAt(cursors, fileId, side, line);
+      if (stopped) {
+        revealLineCursor(stopped, "reveal");
         return "line";
       }
 
@@ -721,6 +736,13 @@ export function useTerminalReview({
       const hunkIndex = file ? reviewHunkIndexForLine(file.metadata.hunks, side, line) : -1;
       if (hunkIndex < 0) {
         return "none";
+      }
+
+      // An empty list means line-level navigation is hidden (`cursor_line = "off"`), not
+      // that the target row is missing: synthesize the line's cursor and reveal it exactly.
+      if (cursors.length === 0) {
+        revealLineCursor(lineCursorAt(cursors, fileId, hunkIndex, { side, line }), "reveal");
+        return "line";
       }
 
       selectHunk(fileId, hunkIndex);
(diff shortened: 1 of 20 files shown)


benvinegar:
@IAMLEIzZ Any reason this is a draft? Should I take a look?

IAMLEIzZ:
> @IAMLEIzZ Any reason this is a draft? Should I take a look?

Sorry for the delay — this week has been extremely busy and I haven't had a chance to review the latest version yet. I'll need to go through it properly before we proceed. I'll remove the draft status once it's ready for review.

src/ui/hooks/useTerminalReview.ts
```diff
@@ -723,6 +738,13 @@ export function useTerminalReview({
         return "none";
       }
 
+      // An empty list means line-level navigation is hidden (`cursor_line = "off"`), not
```
benvinegar, line 741: An empty cursor list doesn’t always mean `cursor_line` is off—it can also happen with extension file views. In that case we may return `"line"` even though there are no bounds to scroll to.

Could we only synthesize the cursor when we know the raw diff is being rendered? Otherwise, we should fall back to the hunk.

src/ui/hooks/useTerminalReview.ts
```diff
@@ -723,6 +738,13 @@ export function useTerminalReview({
...
+      // that the target row is missing: synthesize the line's cursor and reveal it exactly.
+      if (cursors.length === 0) {
+        revealLineCursor(lineCursorAt(cursors, fileId, hunkIndex, { side, line }), "reveal");
```
benvinegar, line 744: This also breaks on zero-count sides. For example, `:lN` on a pure insertion can target an old-side row that doesn’t exist, but we still report a successful jump and allow a comment there.

Could we check that the target has a real rendered line first? A pure insertion/deletion test would be useful too.

benvinegar (commented):
Should we add `:` to the in-app help as well? It feels like a primary navigation shortcut.

PR #812, opened 2026-08-18
fix(ui): keep moved-line colors on wrapped diff rows

HackAttack:
Moved-line coloring never reached anyone with `wrap_lines = true`. Rows have two paint paths, and only the unwrapped one forwarded the move classification to the palette: the wrapped cell builders called `splitCellPalette`/`stackCellPalette` without the `moveKind` argument, so every moved row fell back to the ordinary added or removed tint.

### DIFF
--- a/.changeset/wrapped-moved-line-tint.md
+++ b/.changeset/wrapped-moved-line-tint.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Git's `color.moved` highlights are no longer lost when `wrap_lines` is on.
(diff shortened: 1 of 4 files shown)


benvinegar (approved):
Confirmed the bug locally: moved lines use the correct tint with wrapping off, but fall back to the normal added/removed colors when wrapping is enabled.

This fixes both the stack and split rendering paths. I also verified that the regression tests fail against the old implementation and pass with this change, including after merging into the latest `main`. Looks good to me.

Tiny non-blocking nit: the changeset says `color.moved`, but Git calls this `diff.colorMoved` / `--color-moved`.

PR #813, opened 2026-08-19
feat(diff): syntax highlighting for Bazel/Starlark

HackAttack:
Starlark has no grammar of its own, so map it to Python the way GitHub Linguist does. Reaching the renderer needed two lookup fixes as well: named extensionless files (`pkg/BUILD`, `docker/Dockerfile`) now resolve below the repository root, and the language Hunk resolves is pinned onto the metadata Pierre renders, whose own derivation cannot see mappings keyed by whole filenames.

### DIFF
--- a/.changeset/starlark-highlighting.md
+++ b/.changeset/starlark-highlighting.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": minor
+---
+
+Add syntax highlighting for Bazel/Starlark.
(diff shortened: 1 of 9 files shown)


benvinegar:
I'd prefer if this was done via extensions. Mostly because I'd prefer not to be in the game of deciding what files are recognized for what; I'd prefer if that was promoted upstream in Shiki.

So, I just made it possible to support this case in #851. (It's on main, will go out next minor version, e.g. 0.21.0)

HackAttack:
Reworked on top of main.

HackAttack:
Done in #1124

PR #819, opened 2026-08-19
docs(extensions): list hunk-diff-context extension

astwys:
## Summary

- add `hunk-diff-context` to the Hunk extension directory

 ## Validation

- `bun run website:check`
- `hunk extension install astwys/hunk-diff-context@v0.1.0`
- extension appears in `hunk extension list`
- a remapped `hunk-diff-context.toggle-diff-context` keybinding opens the sidebar


benvinegar:
Oops, already done in https://github.com/modem-dev/hunk/pull/820

I have an automation that sweeps up new extensions and opens a PR automatically

PR #825, opened 2026-08-19
Themes: Fix washed-out diff accents and invisible word highlighting

### LINKED ISSUE #824: Theme derivation guards wash out diff colors and cause invisible highlighting
### What happened?

- connect https://github.com/modem-dev/hunk/pull/806 (closed in favor of this issue and its companion)
- companion to https://github.com/modem-dev/hunk/issues/823

## Problem

Two guard helpers in `src/ui/themes.ts` protect contrast but destroy theme identity in the process.

### 1. `readableDiffSign` rescues with a fixed 45% black/white blend

When a stored accent falls below `MIN_DIFF_SIGN_CONTRAST = 3` against the editor background, the rescue is a single fixed step:

```ts
return relativeLuminance(background) > 0.45
  ? blendHex("#000000", preferred, 0.45)
  : blendHex("#ffffff", preferred, 0.45);
```

All accents get the same 45% wash, desaturating genuine palette colors into pastel/mud. Since sign colors seed the derived row tints and badges, the wash propagates through the whole theme. Themes affected today with *correct* catalog data:

| Theme | Slot | Accent | Contrast | Rescued to |
| -- | -- | -- | -- | -- |
| catppuccin-latte | added | `#40a02b` | 2.96 (misses by 0.04!) | `#235818` |
| catppuccin-latte | modified | `#df8e1d` | 2.31 | `#7b4e10` |
| ayu-light | added / removed | `#6cbf43` / `#ff7383` | 2.23 / 2.55 | `#3b6925` / `#8c3f48` |
| snazzy-light | added / removed | `#2dae58` / `#ff5c57` | 2.77 / 2.93 | `#196030` / `#8c3330` |
| solarized-light | added | `#859900` | 2.97 | `#495400` |
| monokai | removed | `#c4265e` | 2.68 | `#df88a6` (pink) |
| gruvbox dark ×3 | removed | `#cc241d` | ~3.0 / 2.7 / 2.4 | `#e38783` (washed pink) |
| gruvbox light ×3 | modified | `#d79921` | 2.25 / 2.19 / 1.97 | `#765412` |
| min-light | added | `#77cc00` | 2.02 | `#417000` |

(everforest ×2, night-owl, material ×5, tokyo-night, and horizon-bright are also rescued, compounding their catalog data bugs)

### 2. Invisible diff highlighting

`readableTintedBackground` can collapse row and word-emphasis tints to the same color. Row backgrounds use tint 0.12/0.2 and word-level content backgrounds use 0.18/0.28, but the helper steps the tint down by 0.02 until the foreground clears 4.5:1 — so both requests can converge on the identical color, making word-level diff highlighting invisible. Affected today:

| Theme | `addedBg` | `addedContentBg` | Channel delta |
| -- | -- | -- | -- |
| everforest-light | `#eee8d1` | `#eee8d1` | **0** — the removed pair collapses too (`#f2e5d4` == `#f2e5d4`) |
| one-dark-pro | `#3a473d` | `#3a473d` | **0** — word highlighting invisible on added lines |
| material-theme-palenight | `#454f4c` | `#454f4c` | **0** |
| plastic | `#39453b` | `#3b483c` | 6 — near-invisible |

Nothing enforces a minimum separation between row and content tints for any theme.

## Proposed fix

1. **Minimal-adjustment rescue in `readableDiffSign`:** step the black/white blend amount up from a small value until the contrast floor passes (mirroring how `readableTintedBackground` already iterates), so catppuccin-latte's green gets a ~2% nudge instead of a 45% wash. Hue is preserved by construction; saturation loss becomes proportional to how far out of range the accent actually is.
2. **Separation floor for emphasis tints:** after deriving `addedBg`/`addedContentBg` (and the removed pair), require a minimum `hexColorDistance` between row and content backgrounds, nudging the content tint further when the contrast stepping collapsed them.
3. **Table-driven regression tests across all 66 bundled themes:** sign colors pass the guard without rescue whenever the source palette allows it, rescued colors stay within a hue tolerance of the original accent, and row-vs-content emphasis stays above the separation floor.

### Steps to reproduce

`hunk diff` with an affected theme, outlined above

### Expected behavior

_No response_

### Version

0.19.0

masonmcelvain:
- close https://github.com/modem-dev/hunk/issues/824

The two contrast guards in `src/ui/themes.ts` protected readability but destroyed theme identity in the process. This fixes both while keeping every existing contrast invariant.

## Changes

- `readableDiffSign` now rescues low-contrast accents with the smallest black/white blend that clears the 3:1 floor, stepping up from 2%, instead of a fixed 45% wash.
  - Saturation loss is proportional to how far out of range the accent actually is. catppuccin-latte's green (contrast 2.96, missing the floor by 0.04) now gets a ~2% nudge (`#40a02b` → `#3f9d2a`) instead of being crushed to `#235818`.
- Word-emphasis tints (`addedContentBg`/`removedContentBg`) now derive first at full readable strength, and the row tints step further down until the pair clears a 12 channel-distance separation floor (`readableSeparatedRowBackground`).
  - This fixes invisible word-level highlighting on everforest-light, one-dark-pro, and material-theme-palenight (row and content had collapsed to identical colors) and near-invisible highlighting on plastic (distance 6).
  - Pushing content higher would break the readability guarantee that caused the collapse. This is a deliberate deviation from the issue's suggestion.

### Tests

- Table-driven regression tests across all 66 bundled themes: accents that already meet the floor pass through unrescued, rescued accents stay within 2° of the source hue and clear the floor, and row-vs-content separation stays ≥ 12.

## QA

- Spot-check in a real terminal: `--theme catppuccin-latte` (sign colors), `--theme one-dark-pro` or `--theme everforest-light` (word-level highlighting on changed rows).

```
bun run src/main.tsx -- diff --theme catppuccin-latte
```

| Theme | 0.19.0 | Feature |
| -- | -- | -- |
| `catpuccin-latte` (light rescue) | image | image |
| `gruvbox-dark-hard` (dark rescue) | image | image |
| `snazzy-light` (light rescue) | image | image |
| `one-dark-pro` (tint) | image | image |
| `poimandres` (tint) | image | image |
| `everforest-light` (rescue and tint) | image | image |

### DIFF
--- a/.changeset/theme-guard-washout.md
+++ b/.changeset/theme-guard-washout.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Stop the theme contrast guards from washing out diff accents: low-contrast sign colors now get the smallest readable adjustment instead of a fixed 45% blend, and word-level diff emphasis stays visibly separated from row backgrounds.
--- a/.changeset/theme-guard-washout.md
+++ b/.changeset/theme-guard-washout.md
@@ -2,4 +2,4 @@
 "hunkdiff": patch
 ---
 
-Stop the theme contrast guards from washing out diff accents: low-contrast sign colors now get the smallest readable adjustment instead of a fixed 45% blend, and word-level diff emphasis stays visibly separated from row backgrounds.
+Stop the theme contrast guards from washing out diff accents: low-contrast sign colors now get the smallest readable adjustment instead of a fixed 45% blend, and word-level diff emphasis is derived to the renderer's own separation floor so the highlight you see is the one the theme defines.
--- a/src/ui/diff/diffRows.ts
+++ b/src/ui/diff/diffRows.ts
@@ -25,7 +25,7 @@ import type { DiffFile, DiffLineMoveKind } from "../../core/changeset/model";
 import { blendHex, hexColorDistance } from "../lib/color";
 import { measureTextWidth } from "../lib/text";
 import { sanitizeTerminalLine } from "../../lib/terminalText";
-import { TRANSPARENT_BACKGROUND, type AppTheme } from "../themes";
+import { MIN_EMPHASIS_SEPARATION, TRANSPARENT_BACKGROUND, type AppTheme } from "../themes";
 import { expandDiffTabs } from "./codeColumns";
 import type { DiffRow, RenderSpan, SplitLineCell, StackLineCell } from "./diffRowModel";
 import {
@@ -116,7 +116,6 @@ function tabify(text: string, tabWidth: number, initialColumn = 0) {
 // into terminal spans. The same highlighted line objects are reused when files remount or when
 // we build both split and stack rows, so memoize flattened spans by line node + theme/background.
 const flattenedHighlightedLineCache = new WeakMap<HastNode, Map<string, RenderSpan[]>>();
-const MIN_WORD_DIFF_BG_DISTANCE = 28;
 const WORD_DIFF_BLEND_STEP = 0.005;
 const WORD_DIFF_MAX_BLEND = 0.2;
 const wordDiffBackgroundCache = new Map<string, Record<SplitLineCell["kind"], string>>();
@@ -131,7 +130,7 @@ function strengthenWordDiffBg(lineBg: string, signColor: string) {
     const candidate = blendHex(signColor, lineBg, blendRatio);
     strongestCandidate = candidate;
 
-    if (hexColorDistance(candidate, lineBg) >= MIN_WORD_DIFF_BG_DISTANCE) {
+    if (hexColorDistance(candidate, lineBg) >= MIN_EMPHASIS_SEPARATION) {
       return candidate;
     }
   }
@@ -144,8 +143,8 @@ function isHexThemeColor(color: string) {
   return /^#[0-9a-f]{6}$/i.test(color);
 }
 
-/** Resolve one word-diff background without turning transparent surfaces into black blends. */
-function resolveWordDiffHighlightBg(contentBg: string, lineBg: string, signColor: string) {
+/** Strengthen custom-theme overrides whose pair sits too close together. */
+export function resolveWordDiffHighlightBg(contentBg: string, lineBg: string, signColor: string) {
   if (contentBg === TRANSPARENT_BACKGROUND || lineBg === TRANSPARENT_BACKGROUND) {
     return contentBg;
   }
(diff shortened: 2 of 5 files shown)


benvinegar:
Pulled this onto fresh main and dug through it. The incremental sign-color rescue looks right, but I don’t think the PR is ready yet (feel free to push back on any of this).

The word-emphasis part seems to solve the same problem twice. `diffRows.ts` already checks whether a word-level highlight is distinct enough from its row background and strengthens it when needed, using a distance threshold of 28. This PR adds an earlier check in `themes.ts` with a threshold of 12. A color can therefore pass the new theme test and still be changed again by the renderer—77 of 130 current theme pairs do exactly that. Could we keep this policy in one place and test the final rendered highlight instead?

The new sign-color tests also don’t catch the old washout behavior. I replayed the old 45% blend against all 36 rescued catalog accents, and every one still passes the new contrast and hue assertions—blending toward black or white mostly preserves hue. A test for the first passing step, or at least a bound on how far the result can move from the source, would cover the regression better.

One smaller edge case: the rescue chooses white whenever background luminance is at most 0.45, but white can’t reach 3:1 once the background is above 0.30. For example, `#aaaaaa` ends up at about 2.32:1 against white, while black is over 9:1. None of the current bundled themes hit that range, but the helper should still uphold the contrast floor it promises.

I’d keep the incremental sign rescue, tighten its tests, and fold the emphasis fix into the existing renderer guard.

masonmcelvain:
Thanks for the review! I agree with all three points. I updated the PR comment "Feature" screenshots. The following response was written by Claude Fable 5:

One place I'd push back slightly: folding the emphasis fix into the renderer guard goes in the wrong direction. `diffRows.ts` can only strengthen emphasis color and it doesn't check text contrast — so when it rewrites a pair, it can quietly break the theme's 4.5:1 text guarantee. Only the theme layer can create separation by stepping the row tint down while keeping both surfaces readable. So I unified in the mirror direction: bundled themes now derive to the renderer's own floor through a shared constant, which makes the renderer guard a no-op for every bundled theme; the guard stays as a backstop for custom-theme overrides and transparent surfaces.

The test now asserts the final rendered highlight — `resolveWordDiffHighlightBg` must return the theme's pair untouched, with ≥28 separation and ≥4.5:1 text contrast — so the policy is one place and the tested color is the painted color. All 130 pairs still derive readable, visibly tinted rows at 28; the only value that moves is github-dark-default's added row, which steps down one 2% notch.

For the sign-color tests, I added the minimality assertion you suggested: each rescued accent must equal the first 2% blend step (from either anchor) that clears the floor. Replaying the old 45% wash against that fails for all 36 rescued accents.

And `readableDiffSign` now falls back to the opposite anchor whenever the luminance-chosen one can't itself reach 3:1, which closes the (0.30, 0.45] gap; since the fallback only fires when the preferred anchor can't clear the floor, every current bundled theme's output is byte-identical. Added a unit test with your #aaaaaa case.

PR #830, opened 2026-08-21
test(pty): batch repeated scroll input

scttcper:
The horizontal-scroll PTY cases only assert the final clamped state, so send the repeated held-key input as a burst and settle once on the visible result. Discrete arrow behavior stays covered by the in-process interaction suite.

On the passing 114-test PTY partition this took 173.48s to 163.88s, saving 9.60s (5.5%). The layout file itself took 21.82s to 11.57s, saving 10.25s (47%).


scttcper:
Closing this after digging through the test history. #495 intentionally kept the per-key idle wait because firing raw keystrokes can hide dropped or coalesced input. Most of the speedup here comes from removing that guarantee, so this is not the right tradeoff.

PR #842, opened 2026-08-24
docs(website): list hunk-commit-log in the extension directory

sadick254:
## Problem

`hunk-commit-log` reviews a branch one commit at a time: it puts the commit
series in a pane beside the diff, the commit message above it, and binds keys to
step between commits without leaving the window. It is tagged `hunk-extension`
and installable today, but the directory does not list it, so the only way to
find it is the GitHub topic.

## Approach

One entry appended to `EXTENSION_CATALOG` in `website/src/data/extensions.ts`,
in the file's existing alphabetical-by-repository order. Every field is taken
from the extension's own manifest rather than restated: `name` and `version`
from `package.json`, `apiVersion` from its `hunk` block, `repo` exactly as
`hunk extension install` accepts it.

`categories` names the surfaces it registers, using the authoring guide's own
words: two docked panes (the commit list, the commit message), four commands
(`next`, `previous`, `toggle`, `expand`), and a changeset transform that rewrites
the review title to name the reviewed commit and its position in the series.

Non-goals: no change to the page, the schema, or the build-time topic fetch.

## Verification

- `bun test scripts/check-extension-catalog.test.ts` — 9 pass, 0 fail
- `bun run format:check` — clean across 967 files
- `bunx oxlint website/src/data/extensions.ts --deny-warnings` — 0 warnings, 0 errors
- Empty changeset included, since this is a directory listing rather than a
  change to `hunkdiff` behaviour.

The extension itself is verified against Hunk 0.19.0 on macOS: 67 unit tests,
and its panes, mouse row selection, and stepping keys exercised in a live
review.


benvinegar:
@sadick254 Cool!

I do recommend putting some visuals in your repo btw, to help people see what they're getting. The hunk repo has a video making skill file for generating one; ask your agent to find it.

PR #845, opened 2026-08-24
feat(session): navigate to comments by id

### LINKED ISSUE #490: Add --comment flag to hunk session navigate

Extend `hunk session navigate` with a `--comment <comment-id>` selector so the CLI can jump the live viewport to a specific comment by ID, the same way it can already jump by file/hunk/line or by `--next-comment` / `--prev-comment`.

## Why?

Today the `navigate` subcommand only accepts three kinds of targets:

- an absolute position (`--file <path>` + `--hunk <n>` / `--old-line <n>` / `--new-line <n>`)
- the next comment (`--next-comment`)
- the previous comment (`--prev-comment`)

The comment IDs that `hunk session comment list` prints are otherwise opaque: a script can read them, but it has no way to ask the running TUI to focus one. The closest workaround is `navigate --file <path> --new-line <n>` after parsing the comment payload, which is two round-trips and breaks if the comment was anchored to a different side or to a hunk-relative target.

For agent-driven review flows that read the comment list with `--json`, pick the comment they want the human to look at next, and then steer the TUI there, the missing selector is the last hand-off step. A direct `--comment <id>` flag closes the loop and matches the existing ergonomics (`--file` / `--hunk` / `--old-line` / `--new-line` / `--next-comment` / `--prev-comment` are all first-class peers).

## How? (optional)

- Add `.option("--comment <id>", "focus the viewport on the comment with this id")` to the `session navigate` commander setup, alongside the existing positional/option flags.
- In `requireNavigateTarget` (or whichever validator currently enforces "exactly one target"), accept `--comment <id>` as a fourth mutually-exclusive target and reject it when combined with `--file`, `--hunk`, `--old-line`, `--new-line`, `--next-comment`, or `--prev-comment` so the matcher stays unambiguous.
- Resolve `--comment <id>` in the daemon: look up the comment by id (the same path `comment list` already walks), read its `filePath` / `hunkIndex` / `side` / `line`, and emit the same `navigate` payload you'd get from the equivalent `--file --hunk --side --line` triple — no new payload field, no new socket verb, just a server-side resolver.
- Add unit coverage for the commander option (mutual exclusion) and a daemon-level test that `--comment <id>` lands the viewport on the right hunk+side+line.
- Add one `--json` example to `docs/agent-workflows.md` showing the read-comment-then-navigate loop.

Yuki9814:
## Problem

`hunk session comment list --json` exposes stable comment IDs, but callers cannot use one of those IDs to focus that exact comment. They must reconstruct a file/line target or use relative next/previous navigation.

Closes #490.

## Approach

- add `hunk session navigate --comment <id>` and reject combinations with the existing navigation selectors;
- resolve the ID against the selected session's live comments in the daemon, then reuse the existing `navigate_to_hunk` payload;
- return a clear 400 response for an unknown ID and bump daemon compatibility for the new request field;
- document the new mode, regenerate the published references, and add a minor Changeset.

This stays in the existing session CLI/daemon path. It does not add a new UI socket command or change how comments are stored.

## Validation

Tested on macOS arm64:

- `bun ./scripts/run-test-suite.ts` — 3021 passed, 24 platform/tool-gated skips, 0 failed;
- `bun test src/app/cli.test.ts src/session/protocolSchemas.test.ts src/session/agent/cliClient.test.ts src/session/broker/brokerServer.helpers.test.ts src/session/client/capabilities.test.ts src/session/agent/commands.daemon.test.ts src/hunk-review/skillDocument.test.ts scripts/generate-docs.test.ts` — 199 passed, 0 failed;
- `oxfmt --check .`;
- `oxlint . --deny-warnings`;
- `tsc --noEmit`;
- `bun ./scripts/generate-docs.ts --check`;
- `bun ./scripts/generate-changelog.ts --check`;
- `depcruise src packages --config .dependency-cruiser.cjs --ignore-known`.

The Linux-only live-session CLI integration is included but platform-gated locally; CI can exercise it. There is no visual UI change.

Implemented with OpenAI Codex assistance. The final diff and verification results were reviewed before submission.


src/session/broker/brokerServer.ts
```diff
@@ -269,11 +292,13 @@ export async function handleSessionApiRequest(state: HunkSessionBrokerState, req
...
               ...input.selector,
-              filePath: input.filePath,
-              hunkIndex: input.hunkNumber !== undefined ? input.hunkNumber - 1 : undefined,
-              side: input.side,
-              line: input.line,
-              commentDirection: input.commentDirection,
+              filePath: comment?.filePath ?? input.filePath,
+              hunkIndex:
+                comment?.hunkIndex ??
+                (input.hunkNumber !== undefined ? input.hunkNumber - 1 : undefined),
```
benvinegar, line 298: **[P1] Preserve exact-line navigation for comment IDs**

This forwards `hunkIndex` together with `side`/`line`, but `useTerminalReview.navigateToLocation` only calls `revealLine` when `hunkIndex` is absent. As a result, `--comment` selects only the hunk rather than revealing the annotated row, and the command result omits `side`/`line`. The new integration test confirms the regression: `bun test test/session/cli.test.ts -t "navigate works"` fails at `test/session/cli.test.ts:655`. Please omit `hunkIndex` for resolved comments (letting the line resolve its containing hunk), or make the terminal handler prioritize an exact line whenever `side`/`line` are present.

benvinegar (changes requested):
Requesting changes because the new comment-ID navigation path currently drops exact-line reveal semantics and its added integration test fails on Linux.

Validation: `bun test test/session/cli.test.ts -t "navigate works"` (1 failed).

benvinegar (approved):
Verified the exact-line precedence fix. Comment-ID navigation now reaches the annotated row and returns `side`/`line`; the focused unit tests, Linux session integration regression, and typecheck all pass locally. The PR still has a merge conflict with `main` that must be resolved before merging.

benvinegar (approved):
Maintainer-rebased this branch onto current `main` with the contributor's five commits preserved. The only conflict was the recent `parseSessionNavigateCommand` extraction; I retained that structure and applied the PR's `--comment` path within it. Focused tests (206), the live session navigation regression, typecheck, lint, formatting, generated-file checks, and dependency checks pass.

benvinegar:
@Yuki9814 Thanks for the contribution!

PR #850, opened 2026-08-25
feat(cli): add pacman silent install source

marmitar:
No runtime detection for pacman, since different AUR packages may install binaries at different locations (/opt, /usr/bin, /usr/lib). Only viable detection is `pacman -Qo ${executablePath}` or something like `rg ${executablePath} /var/lib/pacman/local/*/files`.

Both cases could affect startup time, so pacman is only available via the `HUNK_INSTALL_SOURCE` environment variable. Either bundled with `--env "HUNK_INSTALL_SOURCE*"` or set at runtime from a custom wrapper.

Because of how pacman packages are managed, no update notices or specific update instructions should be shown.

### DIFF
--- a/.changeset/olive-impalas-nail.md
+++ b/.changeset/olive-impalas-nail.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Recognize pacman/AUR-managed installations via `HUNK_INSTALL_SOURCE=pacman` and suppress automatic update notices for them.
--- a/src/core/process/updateNotice.ts
+++ b/src/core/process/updateNotice.ts
@@ -34,8 +34,10 @@ export type { InstallSource, UpdateChannel };
  * starts. A notice there would ask the user to fix something mise just fixed. A local source build
  * is replaced by rebuilding the checkout it came from, which is the developer's own workflow and
  * not something a published version number should interrupt.
+ *
+ * Arch packages installed through pacman or an AUR helper are updated externally.
  */
-const SILENT_INSTALL_SOURCES: readonly InstallSource[] = ["mise", "dev"];
+const SILENT_INSTALL_SOURCES: readonly InstallSource[] = ["mise", "pacman", "dev"];
 
 export interface UpdateNoticeDeps {
   env?: NodeJS.ProcessEnv;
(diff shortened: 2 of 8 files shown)


marmitar:
Sorry, forgot to mention. I maintain the AUR package [hunk](https://aur.archlinux.org/packages/hunk) and I plan on using it there. I'm using `HUNK_INSTALL_SOURCE=mise`, but the message don't really make sense there.

marmitar:
@benvinegar could you please review this? Or implement an equivalent feature?

benvinegar:
Yep, taking a look, just wrapping my head around what it means to have 2 packages here:

image

marmitar:
AUR packages are usually built from source, that's what the `hunk` package does. In contrast, `hunk-bin`  just fetches the pre-built binary from the release page.

The un-updatable `pacman` install source would make sense for both packages in this case.

src/core/process/updateNotice.ts
```diff
@@ -34,8 +34,10 @@ export type { InstallSource, UpdateChannel };
  * starts. A notice there would ask the user to fix something mise just fixed. A local source build
  * is replaced by rebuilding the checkout it came from, which is the developer's own workflow and
  * not something a published version number should interrupt.
+ *
+ * AUR packages installed via pacman should be updated manually or via an AUR helper.
```
benvinegar: Minor wording nit: AUR packages aren’t installed by pacman directly. Could this say “Arch packages installed through pacman or an AUR helper” instead?

.changeset/olive-impalas-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Support pacman/AUR installs via `HUNK_INSTALL_SOURCE=pacman`.
```
benvinegar: Could we clarify that this recognizes pacman/AUR-managed installs rather than adding an installation method? Maybe: “Recognize pacman/AUR-managed installations via `HUNK_INSTALL_SOURCE=pacman` and suppress automatic update notices for them.”

src/core/process/updateNotice.ts
marmitar: Done.

.changeset/olive-impalas-nail.md
marmitar: Done.

PR #865, opened 2026-08-26
feat(ui): make file and hunk spacing configurable

### LINKED ISSUE #839: Make vertical space between files/hunks configurable

Add a way to make vertical spacing between files primarily, and ideally smaller spaces between hunks within the same file.

### Why?

It would be much easier to see diff boundaries at a glance.

### How? (optional)

Perhaps specifying the number of lines to make a gap in the config.

Ajay-Satish-01:
## Summary
- Adds `file_gap` / `--file-gap` (0–8, default `1`) so the space between files in the review stream is configurable. `1` keeps the current `─` rule, `0` hides it, and larger values add blank rows above the rule.
- Adds `hunk_gap` / `--hunk-gap` (0–8, default `0`) so later hunks can have blank rows before `@@` (still inserted when hunk headers are hidden).
- This is presentation-only in the TUI/OpenTUI stream (not shared review geometry, View-menu persistable, static pager, or browser CSS).

Fixes #839

## Test plan
- [X] `bun run start -- diff --no-sidebar --mode stack --file-gap 3 --hunk-gap 2` on a repo with at least two files and two hunks
- [X] `--file-gap 0` hides the `─`; `--file-gap 1` is the current look
- [X] `--hunk-gap 2` puts two blanks before later `@@`; first file/hunk stay flush
- [X] `--no-hunk-headers --hunk-gap 2` still inserts blanks before later hunk content
- [X] Config `file_gap` / `hunk_gap` apply; CLI flags override; `--file-gap 9` and `--hunk-gap abc` error
- [X] `[` / `]` still land on hunk content, not the spacer

Verified: unit tests for parse/config/plan/geometry; PTY layout tests for `--file-gap 3` and `--hunk-gap 2`; live PTY frames for 0/1/3 file gaps and 0/2 hunk gaps (including hidden headers). `format:check`, `lint`, `typecheck`, `check:docs`, `deps:check`, and `bun run test` passed.

## Screenshots

Real Hunk TUI (PTY capture), 100×32, `--no-sidebar --mode stack`, default theme, macOS.

### Default look (`file_gap=1`, `hunk_gap=0`)
One `─` rule between files; no extra space before later hunks.

Default file and hunk spacing

### `--file-gap 0`
Hides the `─`; the next file header sits on the previous body.

file-gap 0

### `--file-gap 3`
Two blank rows, then the `─` rule, then `beta.ts`.

file-gap 3

### `--hunk-gap 2`
Two blank rows before the later `@@` in `alpha.ts`. First hunk stays flush.

hunk-gap 2

### `--file-gap 3 --hunk-gap 2`
Both gaps together.

file-gap 3 and hunk-gap 2

### `--no-hunk-headers --hunk-gap 2`
Blanks still appear before later hunk content when hunk headers are hidden.

no hunk headers with hunk-gap 2


Ajay-Satish-01:
@benvinegar, would you be able to take a look at this? Thanks

benvinegar:
I'm good with the behavior (adding the config + feature) - good call

I just want to look over the code first

Ajay-Satish-01:
> I'm good with the behavior (adding the config + feature) - good call
> 
> I just want to look over the code first

Sure. Thank you

benvinegar (approved):

PR #867, opened 2026-08-27
fix(core): index moved hunk lines that start with ++ or --

benvinegar:
## Before / after

Real terminal captures — `tmux` panes running Hunk's interactive TUI on the same `git diff --color-moved=zebra` input, buggy code (origin/main) vs. this fix.

| BEFORE | AFTER |
| :---: | :---: |
| before | after |

The moved tint glows on the `++ plus` row before the fix (it was silently dropped from the indexer), and the real moved line `+ AAA one…` renders as a plain addition. After the fix, the tint sits exactly on the moved line.

---

The hunk adds `++ plus` right before a moved line. The classifier indexes every `+`/`-` row it sees; the `++ plus` row is one of them — but its diff body reads `+++`, so the old guard treated it as a file header and skipped it, shifting every later classification by one row.

## Root cause

`collectLineMoveKinds` guarded hunk-body `+`/`-` lines against looking like `+++ b/…` / `--- a/…` file headers, but those headers never appear inside a hunk. A content line whose own text starts with `++` (e.g. `++i;`, `++ plus`) reads as `+++…` in the patch and was skipped, so:

- a genuine `--color-moved` line lost its "moved" styling, and
- one other row gained a moved badge it didn't earn.

## Fix

Removed both guards (`src/core/changeset/fromPatch.ts`): inside a hunk, every line starting with `+` is an added content row and every line starting with `-` is a deleted content row; `inHunk` already excludes the actual headers.

## Verification

- New regression tests in `src/core/changeset/fromPatch.test.ts` for both sides (fail on the old code, pass on the fix).
- End-to-end check against a real zebra patch: parser output matches ground-truth move indices exactly (was off by one).
- `bun test src/core/changeset` + `bun run typecheck` green.


benvinegar:
Real terminal captures (tmux panes running the interactive TUI, buggy vs. fixed): https://sideshow.sh/p/2VKXeBS2W-E

benvinegar:
The real terminal captures are inline in the PR body above (before/after side by side).

PR #920, opened 2026-08-30
fix: store persistent state under XDG state directory

### LINKED ISSUE #821: `state.json` (and extensions dir) hardcoded under `$XDG_CONFIG_HOME` instead of `$XDG_STATE_HOME`
### What happened?

hunk stores its state file in `$XDG_CONFIG_HOME/hunk/state.json` right alongside its config file `$XDG_CONFIG_HOME/hunk/config.toml`.

Per the [XDG Base Directory spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), `$XDG_CONFIG_HOME` is for user-editable configuration, while `$XDG_STATE_HOME` (defaulting to `~/.local/state`) is for data that "should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in `$XDG_DATA_HOME`" — which, seems to me, describes `state.json`'s contents exactly.

### Steps to reproduce

Open `$XDG_CONFIG_HOME/hunk` and find `state.json` inside.

### Expected behavior

`state.json`should live in `$XDG_STATE_HOME/hunk`.

### Version

0.19.0

mikemikimike:
## Problem

Hunk currently stores its persistent state file at `$XDG_CONFIG_HOME/hunk/state.json`, alongside the user-editable configuration at `$XDG_CONFIG_HOME/hunk/config.toml`.

This mixes two different kinds of data defined by the XDG Base Directory Specification: `XDG_CONFIG_HOME` is intended for user-editable configuration, while `XDG_STATE_HOME` is intended for data that persists across restarts but is not important or portable enough to belong in `XDG_DATA_HOME`. The state file and the user-scoped directory used for globally installed extensions belong in the latter category.

## Changes

- Resolve `state.json` from `$XDG_STATE_HOME/hunk`, falling back to `~/.local/state/hunk` when `XDG_STATE_HOME` is unset.
- Keep `config.toml` under `$XDG_CONFIG_HOME/hunk`, with the existing `~/.config/hunk` fallback.
- Move global extension discovery and managed extension installs to the same XDG state directory.
- Add regression coverage for explicit XDG directories and state-directory fallbacks.

## Validation

- `bun test src/core/run/paths.test.ts src/extensions/discovery.test.ts src/extensions/startup.test.ts src/extensions/manage/install.test.ts`
- `bun run typecheck`
- `bun run lint`
- `bun run deps:check`
- Changed files pass the formatter check.
- `git diff --check`

The repository-wide format check reports pre-existing failures elsewhere in the checkout; the changed files pass the formatter check.

## Compatibility / Known limitations

This changes the default location of persistent state and globally managed extensions on XDG-compatible systems. User-editable configuration remains at its existing location. Existing state or extensions under the old configuration directory are not migrated automatically.

Fixes #821


benvinegar:
Can you fix the PR description? It's poorly formatted. What problem are you solving, etc? (I can guess, but ...)

mikemikimike:
I rechecked this PR for issue #821. The PR description covers the reported persistent-state behavior, and I do not see a human changes-requested review or a failing check for the current head. Could a maintainer please review whether this fully resolves the issue and approve it, or advise on any remaining work?

mikemikimike:
The PR description now documents the problem, the XDG state/config split, the compatibility limitation, and the focused validation. Issue #821 is still open. Please review the current head when available; Vercel authorization is intentionally ignored.

PR #925, opened 2026-08-30
feat(review): add threaded inline comments

benvinegar:
## Problem

Inline reviewer notes could only be created and deleted. Reviewers could not correct a saved note or continue a discussion beside the affected code, which forced context into disconnected notes or external review tools.

## Approach

- add editable user notes and arbitrarily nested replies to the shared review model, intent planner, reducer, selectors, and protocol projections
- preserve note identity and anchor metadata during edits, inherit anchors for replies, and reject invalid/blank mutations or parent removal that would orphan descendants
- render connected inline threads in the terminal with bounded visual indentation, exact shared geometry, compact file/range titles, and mouse/keyboard Reply, Edit, Delete, Save, and Cancel actions
- preserve the viewport when mouse-opening Edit or Reply while retaining keyboard reveal behavior
- publish parent identity and edit lifecycle data through the existing session, broker, conformance, and extension snapshot surfaces; bump the extension API to v12

This belongs in review core because thread identity, mutation rules, ordering, and cross-surface parity are shared review semantics. Rendering, mouse hit areas, composer focus, and viewport preservation remain terminal-local.

## Important boundaries

- only editable reviewer-authored notes can be edited
- replies currently target semantically stored live/user notes, not static sidecars without durable identities
- logical nesting is unlimited; terminal indentation is visually bounded
- global `stateRevision` remains the concurrency control

## Verification

- `bun run format`
- `bun run typecheck`
- `bun run lint`
- `bun run deps:check`
- source-boundary and review-vocabulary suites: 18 passed
- `bun run test`: 2,067 passed, 10 skipped
- `bun run test:integration`: 136 passed, 1 platform skip
- post-polish focused AgentInlineNote tests: 19 passed
- post-polish `bun test test/pty/notes.test.ts`: 19 passed
- `bun run test:tty-smoke`: 9 passed
- `bun run install:bin`
- installed-binary real-TTY transcript smoke: exited 0
- `git diff --check`

## Platforms

Tested on Linux. macOS and Windows were not tested locally; the existing macOS-only PTY lifecycle test remained skipped.

## Visual evidence

Validated in the real Hunk TUI through PTY interaction coverage and a local installed-binary TTY transcript smoke. No media is attached.

## Release

Includes a minor `hunkdiff` Changeset.


lekemula:
Hi @benvinegar, thanks for adding this great missing feature! 

I checked it out; however, I noticed an issue entering edit mode using the keyboard only. When navigating with `j/k` toward the saved note, the only way to enter in edit note seem to only be possible via "mouse":


https://github.com/user-attachments/assets/f717e5d6-de87-4dfb-b913-d20dff152dc7

As seen on the recording above, even when I hover with the mouse & press `e`, it goes into editing the file and not the saved note. 

I'm using iTerm + tmux - wonder if this has anything to do with it.

Am I missing something? Let me know if I can provide you with more information.

PR #962, opened 2026-08-31
feat(ui): let note composer save be remapped

### LINKED ISSUE #846: Allow the note composer save note command (Ctrl-S) to be remapped

Let the note composer's save key come from the `[keybindings]` table instead of
being hardcoded to Ctrl-S. Something like a `hunk.review.saveNote` command id
that the composer actually consults, shipping with `ctrl+s` as the default so
nothing changes for anyone who doesn't touch it.

`docs/keybindings.md` currently says the opposite:

> Keys that belong to a dialog, menu, or focused text input — `Esc`, `Enter`,
> `Ctrl-S` while writing a note — are part of those widgets rather than
> commands, and are not remappable.

But I don't see why.

### Why?

Short version: under Zellij, Ctrl-S never makes it to Hunk at all. Zellij's
stock keybindings grab it in every mode except Locked:

```kdl
shared_except "scroll" "locked" {
    bind "Ctrl s" { SwitchToMode "Scroll"; }
}
```

I type up a note, hit Ctrl-S, and instead of saving, the pane drops into
Zellij's scroll mode and my draft just sits there unsaved. Every other key I
touch during a review is fine, and every other Hunk command I'd want to move is
already remappable — the save chord is the one thing I can't route around from
Hunk's side.

Repro:

1. `zellij` with stock keybindings (no `clear-defaults=true`)
2. `hunk diff` in a pane, `c` to start a note, type whatever
3. Ctrl-S → Zellij switches to scroll mode, draft never saves

Setup: Hunk 0.18.1 (Homebrew), Zellij 0.44.3, iTerm2, zsh, macOS 26.3,
`TERM=xterm-256color`.

**How this is different from what's already been filed:** #345 was this exact
symptom under tmux, and #350 fixed it by widening `isSaveDraftNoteKey` to catch
more *encodings* of Ctrl-S — raw `\x13`, CSI-u `ESC[115;5u`, tmux control mode.
That trick can't save me here, unfortunately: Zellij eats the keystroke before
Hunk's stdin ever sees it, so there's no encoding left to normalize. And #491
shipped user keybindings via #611, but that pass scoped focused-input keys out
on purpose. 

**Could this be an extension?** I went looking, and I don't think so. Per the
routing precedence in `docs/keybindings.md`, focused text inputs sit *above*
session extension keyboard modes, so an extension mode can't claim a chord while
the composer has focus — and the composer's save handler isn't reachable through
`hunk.executeCommand` either. Seems like it has to be a host change.

### How? (optional)

Narrow version: register save as a real command spec in `builtinCommandSpecs`
with `defaultKeys: ["ctrl+s"]`, then have `handleFocusedInputShortcut` check the
*resolved* chord for that id instead of calling `isSaveDraftNoteKey`. All the
encoding-normalization work from #350 stays exactly where it is — it just
becomes part of how `ctrl+s` gets matched rather than a special case that only
saving knows about.

Focused inputs still resolve before the command table, so the rule survives;
only the chord becomes configurable. Then the composer footer hint
and the `?` dialog would show the effective key for free, same way #614 wired up
the rest of the surface.

Ajay-Satish-01:
## Summary

- Adds `hunk.review.saveNote` (default `ctrl+s`) so the note composer save shortcut comes from `[keybindings]` instead of a hardcoded widget key.
- Escape, Enter, and F10 stay widget-owned. While the composer is focused, only the **resolved** chords save; unbinding or remapping away from `ctrl+s` also stops CSI-u / raw Ctrl-S.
- The composer footer and `?` help show the live chord (`Ctrl+S` by default, e.g. `Ctrl+Enter` after a remap). Unbound omits the key and leaves a clickable Save.
- This has to be a host change: focused text inputs sit above session extension keyboard modes, so an extension cannot claim the chord while the composer has focus.

Fixes #846

## Test plan

- `c` to start a draft; footer shows `Ctrl+S save`; Ctrl-S (including CSI-u `\e[115;5u`) saves
- Config `"hunk.review.saveNote" = "ctrl+enter"`: footer shows `Ctrl+Enter save`; Ctrl-S does not save; Ctrl+Enter does
- Config `"hunk.review.saveNote" = false`: Ctrl-S and CSI-u leave the draft; `?` drops the save-note row
- Clicking `Ctrl+S save` still saves
- `command_executed` fires with `hunk.review.saveNote` on a successful save

Verified: unit tests for catalog, chord matching (including Ctrl+Shift+S no longer saving), execute-by-id ownership, help remap/unbind; AppHost remap to `ctrl+enter`, unbind, and CSI-u after remap-away; `bun test ./test/pty/notes.test.ts` (19 pass, including live `Ctrl+S save` and click-to-save). `format:check` and lint on the changed files passed. `bun run test` had four failures already present on this machine (Jujutsu source logging, two Shiki catalog checks, install VM symlink contract); none of those files are in this diff.

## Screenshots

Composer footer is driven by the resolved `hunk.review.saveNote` label (default `Ctrl+S save Esc cancel`). After `"hunk.review.saveNote" = "ctrl+enter"`, AppHost frames show `Ctrl+Enter save`. I can attach live TUI captures if useful.


Ajay-Satish-01:
@benvinegar, would you be able to take a look at this? Thanks

benvinegar:
Good find. I'll probably queue up for the next minor release.

Ajay-Satish-01:
> Good find. I'll probably queue up for the next minor release.

Sure. Thanks

benvinegar:
Still works after the recent repo move, but this command should only be enabled while a draft exists. Right now its key is consumed globally and emits `command_executed` even when there’s nothing to save. Otherwise the approach looks good.

Ajay-Satish-01:
> Still works after the recent repo move, but this command should only be enabled while a draft exists. Right now its key is consumed globally and emits `command_executed` even when there’s nothing to save. Otherwise the approach looks good.

I worked on the change. Please check and let me know if this approach works.

PR #975, opened 2026-09-03
fix: keep one unparseable file from zeroing the whole review

vkurilyak:
I tried [Hunk](https://github.com/modem-dev/hunk) after reading https://x.com/mitchellh/status/2052128048288567617 by Mitchell Hashimoto (https://github.com/mitchellh).

I need better tooling for code reviews before merging changes into production (main branches on GitHub), so a terminal review tool for changesets is exactly what I was looking for.

Hunk showed `0 files +0 -0` and `No files match the current filter.` on a working copy with 2275 changed files, while a scoped `hunk diff -- <single-file>` worked fine.

Root cause was 2 PDF files in the same review, both emitted by the VCS as text rather than `Binary files ... differ`:
- one is a DocuSign PDF
- one is a PDF with lots of images

Both contain `ESC ] ... BEL`-looking binary bytes spanning newlines. `stripTerminalControl` ate across the newlines, corrupting `@@` hunk line counts, so `@pierre/diffs` threw `hunk line count mismatch` / `hunk has more lines than expected` and `changesetFromPatch` returned zero files for the whole review.

This change limits DCS/OSC stripping to the same line so binary-looking bytes cannot eat newlines, and adds a per-file-chunk fallback so one unparseable file becomes a visible placeholder instead of discarding all files. It also adds regression coverage for both the newline guarantee and the keep-good-files fallback.

Tests:

- `bun test src/core/patch src/core/changeset`
- verified Hunk's real `changesetFromPatch` on the original 6.1 MB patch: 2275 files (was 0)

### DIFF
--- a/src/core/changeset/fromPatch.test.ts
+++ b/src/core/changeset/fromPatch.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, test } from "bun:test";
+import { changesetFromPatch } from "./fromPatch";
+
+const goodFile = [
+  "diff --git a/notes.txt b/notes.txt",
+  "new file mode 100644",
+  "index 0000000000..3b18e512d6",
+  "--- /dev/null",
+  "+++ b/notes.txt",
+  "@@ -0,0 +1,2 @@",
+  "+hello",
+  "+world",
+  "",
+].join("\n");
+
+// A hunk body whose line counts disagree with its header: unparseable, but it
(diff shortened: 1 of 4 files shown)


benvinegar:
@vkurilyak Thanks for fixing this!

benvinegar:
Kind of need this to pass CI before I can merge.

PR #996, opened 2026-09-06
fix(ui): keep suspended jobs alive upon fg

HackAttack:
When backgrounding with Ctrl-Z in the shell, `fg` can kill the process instead of bringing it back to the foreground.

Before stopping Hunk, OpenTUI prepares the terminal for the shell: it leaves raw mode, stops reading stdin, and removes its keep-alive timer. The stopped process remains alive, but after `fg` wakes it, the runtime
could exit before Hunk's SIGCONT callback restarted the renderer.

The stop actually takes effect before kill returns, since POSIX delivers a signal sent to the caller's own process group before the call completes. So suspend and resume become one straight line, and staying on that call stack is what keeps the job alive: the event loop never gets a chance to go idle and exit. That retires the SIGCONT listener along with the per-suspend state tracking it needed.

It also closes a hang. The old fallback only ran when kill threw, so a SIGTSTP that was delivered and discarded — as POSIX requires for an orphaned process group — left the app suspended forever, waiting for a SIGCONT nobody would send. Both cases now reach the same restore.

### DIFF
--- a/.changeset/steady-jobs-resume.md
+++ b/.changeset/steady-jobs-resume.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Keep suspended Hunk jobs alive so `fg` restores the TUI and its in-progress state.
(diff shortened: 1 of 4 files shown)


benvinegar:
Thanks!

PR #1048, opened 2026-09-08
fix(ui): save note drafts without stale input races

benvinegar:
## Summary

- scope textarea updates to the draft instance that emitted them
- flush the focused editor value before Ctrl-S or mouse saves consume the draft
- keep draft identities unique when notes are reopened within one millisecond
- cover stale callbacks and fast-save behavior at unit and PTY levels

## Stack

- depends on #1047 so required Knip checks are green while `main` still contains the unused export
- this PR diff contains only the note race fix

## Validation

- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run deps:check`
- `bun run knip`
- `bun run changeset:status`
- `bun run test` (2,100 passed)
- `bun run test:integration` (163 passed, 1 platform skip)
- `bun run test:tty-smoke` (10 passed)
- fast-save PTY regression repeated 10 times

### DIFF
--- a/packages/hunk/src/ui/hooks/useUserNoteComposer.test.tsx
+++ b/packages/hunk/src/ui/hooks/useUserNoteComposer.test.tsx
@@ -398,6 +398,33 @@ describe("useUserNoteComposer", () => {
     }
   });
 
+  test("does not save when the editor belongs to a stale draft", async () => {
+    const updates: Array<[string, string | undefined]> = [];
+    let saveCount = 0;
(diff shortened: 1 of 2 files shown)


benvinegar:
Rerouting this change through #1050, which which? The stack reversal marked this PR merged as merged only because its commit became an ancestor of the former branch cleanup branch; it was not merged to main. #1050 is the active note-r4 PR against main.

PR #1053, opened 2026-09-08
feat(extensions): add host-owned file-view syntax highlighting

### LINKED ISSUE #1023: Let file views show source lines with hunk's syntax highlighting

Give extension file views a way to render source text with the same syntax highlighting the raw
diff gets.

Today a file-view row is a list of symbolic spans (`ExtensionFileViewSpan`), and each span can only
carry one of six semantic tones: `muted`, `accent`, `accent-muted`, `syntax`, `added`, `removed`
(`src/extension-api/types.ts`, `docs/extensions.md` "Layouts use an omitted tone for ordinary text
and generic symbolic tones…"). `syntax` resolves to one flat color, `theme.syntaxColors.default`
(`src/ui/components/panes/FileView.tsx`, `fileViewToneColor`). So a view that presents real source
lines, for example a whole file with unlimited context, renders every token in the same color, while
the raw diff next to it is highlighted through the Shiki worker.

Proposed shape, smallest first:

1. A per-row opt-in, `highlight: { side: "old" | "new", line: number }` or reusing the existing
   `sourceRanges` entry when the row text equals that source line. Hunk then paints the span with
   the file's language through the existing highlight worker, exactly like a raw diff row, and
   layers the row's `tone` (added/removed background or tint) on top.
2. Alternatively a row-level `language?: string` plus a `code: true` flag for rows whose text is
   arbitrary source, highlighted at paint time and cached by `(text, language, theme)`.

Layout stays deterministic and theme-independent either way: highlighting only changes colors at
paint time, never measurement, which is the invariant the current design protects.

## Why?

Extensions that present code rather than a rendered document lose highlighting the moment they
take over a file. The hunk-viewed extension (https://github.com/jacegodk/hunk-viewed) has a
"Full file" view that rebuilds the whole file as a unified or side-by-side diff with unlimited
context, mapped back to source through `sourceRanges`. It works, but reading 2,000 monochrome
lines next to a highlighted raw diff is a clear step down, and there is no way to close that gap
from the extension side: tokenizing in the extension does not help, because spans cannot carry
per-token colors, only the six tones and bold/italic/underline/strikethrough.

The same applies to any view that shows source: a blame-style view, a "show N more lines" context
expander, a side-by-side rename view, or a snippet inside a rendered-markdown code fence.

## How? (optional)

- `sourceRanges` already tells hunk which exact source line a row presents, and the host already
  reads that side to validate the mapping. Option 1 reuses that: when a row has a single-line
  `sourceRanges` entry and an opt-in flag, `FileView` asks the highlight worker for that line of
  the file (same language detection as the raw diff, same cache) and paints the span from the
  returned segments instead of `fileViewToneColor`. Row `tone` still applies as today, so an
  `added` row keeps its green treatment with highlighted tokens inside.
- The worker request can be scoped to the rows in the render window, as the raw diff does, so a
  10,000-row view does not highlight eagerly.
- No layout API change: rows measure exactly as now, and a view that does not opt in is unchanged.
- API version bump with the new optional field; the folded/full views in hunk-viewed would adopt
  it immediately and can serve as a test bed.

Happy to open a PR if the shape is acceptable.

benvinegar:
## Summary

Closes #1023.

File-view extensions can now declare complete code documents and map symbolic spans to exact UTF-16 source slices. Hunk retains ownership of language detection, Shiki/Pierre tokenization, active-theme colors, worker scheduling, caching, terminal safety, projection, and plain-text fallback.

This enables source, generated-code, blame, Markdown-fence, guttered, and split old/new presentations without exposing token colors or renderer internals to extensions.

## Approach

- Add the import-free API-v28 `codeDocuments` and `span.syntax` contract with bounded validation, normalization, immutable snapshots, and exact text matching.
- Generalize compact highlighting artifacts and the worker protocol for complete document jobs.
- Add one bounded shared document-highlight service with strong identities, single-flight work, subscriber-local cancellation, retry classification, and byte/entry-bounded caches.
- Migrate expanded-source highlighting to that shared service.
- Demand FileView documents only for mounted/halo rows, then project token ranges over retained authoritative span text.
- Preserve native OpenTUI word wrapping and fixed two-cell tabs; require native measurement so syntax paint cannot alter geometry.
- Add API documentation, a checked-in old/new code-document example, package-consumer coverage, and a minor `hunkdiff` Changeset.

## Important invariants and non-goals

- Complete documents are tokenized so multiline lexical state remains correct.
- Highlighting is paint-only: it cannot change text, wrapping, row height, notes, navigation, selection backgrounds, layout generation, or scrolling.
- Unsupported languages, limits, cancellation, stale results, and failures retain ordinary symbolic FileView content.
- `sourceRanges` remain note/navigation bindings and are independent from syntax references.
- Extensions never provide colors or access Shiki, Pierre, HAST, workers, or terminal renderer types.

## Validation

Passed after rebasing onto current `main`:

- `bun run typecheck`
- `bun run lint`
- `bun run deps:check`
- `bun run check:pack`
- `bun run check:docs`
- `bun run website:check`
- `bun run changeset:status`
- `bun run test:integration` — 183 passed, 1 platform skip
- `bun run test:tty-smoke` — 10 passed
- `git diff --check origin/main...HEAD`

Focused unit, React lifecycle, service/worker, compiled-worker, consumer, and FileView projection/geometry suites also passed throughout implementation. The new file-view PTY case verifies real syntax colors, multiline offscreen lexical state, theme changes, resize, refresh generations, cross-file navigation, and exact viewport stability.

`bun run test` retains one environment-dependent failure in unchanged `packages/hunk/src/extensions/hostRuntimeModules.test.ts:179`: this machine has an external `/tmp/node_modules/react`, so the test's deliberately outside-extension import resolves. The failing file is unchanged from `main`; the two shards otherwise reported 1,882 and 2,379 passing tests, respectively.

## Platform and visual evidence

- Validated on Linux with Bun 1.4.2 and Node.js 22.
- macOS and Windows compiled-worker execution were not performed.
- Real PTY and TTY smoke coverage exercised the terminal UI. No standalone screenshot/video is attached; the deterministic PTY test covers syntax paint and layout stability across theme, resize, refresh, and file switching.

## Follow-up adopter

A reviewed `hunk-viewed` migration exists locally and uses complete old/new documents for its unified and split full-file views. Its release is intentionally sequenced after the next minor `hunkdiff` release publishes API v28.


jacegodk:
Tested #1053 with hunk-viewed's full-file view (unified and split, old and new documents, search spans interleaved): tokenization, exact-text validation and fallback all behave as documented. Thanks.

One gap for diff-presenting views: a token color always overrides `tone`, and file-view rows get only `theme.selectedHunk` or `theme.panel` as background. So once an added or removed row is syntax painted, nothing but the `+`/`-` marker distinguishes it from context. In the raw diff the row background carries that signal. Could a row (or span) opt into the raw diff's added/removed background, e.g. `background: "added" | "removed"`, painted under the token colors? That is the "layer the tone on top" part of #1023. Until then hunk-viewed highlights context lines only and keeps changed lines in solid tone.

Two smaller points: a single span/text mismatch rejects the whole layout, so extensions must re-verify every reference; dropping `syntax` on that span would be more forgiving. And the 10,000-line cap is aggregate across documents, so an old+new view fits files up to about 5,000 lines while the row cap allows 10,000.

PR #1057, opened 2026-09-08
perf(test): make suite sharding configurable

benvinegar:
## Problem

Hunk's test runner can shard only its default Linux suite. Other suites and CI platforms cannot reuse the independent-process supervision without duplicating it, while naïvely sharding filtered or coverage runs produces incorrect failures or output.

## Approach

- add named test groups to the existing independent-process shard runner
- route `test:integration` through the same supervised runner
- allow CI to explicitly enable measured shard counts on non-Linux platforms
- keep resource-intensive PTY tests and non-Linux developer runs serial by default
- force filtered and coverage-producing invocations to remain serial so zero-match shards cannot fail focused runs and shards cannot clobber shared coverage output

This keeps Bun's isolated `--parallel` mode disabled because it is incompatible with OpenTUI's native FFI initialization.

## Validation

- `./node_modules/.bin/bun test scripts/test/run-test-suite.test.ts`
- `./node_modules/.bin/bun run test`
- `./node_modules/.bin/bun run test:integration`
- focused integration invocation with `--test-name-pattern`
- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run deps:check`

An explicit two-shard PTY benchmark completed locally, but a hosted-runner responsiveness test exposed resource contention. PTY therefore remains serial by default; explicit sharding is available only for measured environments. Windows command construction is unit-tested here and exercised by the next PR.

## Stack

1. **#1057 — configurable test sharding** (this PR)
2. #1058 — parallel CI validation lanes
3. #1059 — faster Nix validation

Merge in this order.

*This PR description was generated by Pi using OpenAI GPT-5.6 Sol*


benvinegar:
Force-pushed the stack after CI showed that concurrent PTY processes can starve a responsiveness test on hosted runners. PTY tests now stay serial by default; the runner still supports explicit measured shard counts and keeps filtered/coverage runs safe.

benvinegar:
Fixed the file-backed reporter finding: `--reporter-outfile` now forces serial execution just like coverage and filtered runs, covering both `--reporter-outfile=path` and split argument forms in the unit test. The stack was cascade-rebased and pushed.

Validation: Bun 1.4.2 runner tests (9 passed, 33 assertions), typecheck, and lint.

PR #1080, opened 2026-09-09
fix(extensions): reject non-callable VCS operation watch hooks

### LINKED ISSUE #763: Reject malformed optional VCS operation hooks during extension registration
## Problem

Extension VCS registration validates that an operation's `load` field is callable, but does not validate the optional `watchSignature` and `watchPlan` hooks.

An untyped extension can therefore register an operation such as:

```ts
{
  load: async () => result,
  watchSignature: "not a function",
}
```

Registration accepts the operation. Watch planning later attempts to invoke the truthy value and fails with a delayed `TypeError` rather than reporting a registration problem.

The current conversion also widens every key-specific operation through `ExtensionVcsOperation<VcsReviewInput>`, erasing the relationship between the operation key and its input type.

## Proposed direction

- Require `watchSignature` and `watchPlan` to be absent or callable during registration.
- Normalize only the known VCS operation keys into `VcsOperations`.
- Preserve each key's input type through an explicit switch or key-preserving converter rather than widening handlers to the full input union.
- Continue dropping malformed operations so callers receive the existing unsupported-operation behavior.

## Acceptance criteria

- Registration rejects or drops an operation with a non-callable `watchSignature`.
- Registration rejects or drops an operation with a non-callable `watchPlan`.
- Valid optional hooks continue to run through the existing error normalization.
- Unknown operation keys do not enter the internal operation map.
- Add adversarial extension robustness tests for malformed optional hooks.

edenbuilds:
## Summary

`registerVcsAdapter` validates that an operation's `load` field is callable, but not the optional `watchSignature` / `watchPlan` hooks. An untyped extension can register:

```ts
{
  load: async () => result,
  watchSignature: "not a function",
}
```

Registration currently accepts this. The `watchSignature && {...}` spread guard in `toInternalVcsOperation` treats any truthy value as present, wraps it, and the resulting `TypeError` only surfaces once watch planning invokes the hook — well after registration.

## Fix

`toInternalVcsAdapter` now requires `watchSignature` and `watchPlan` to be absent or callable before accepting an operation, matching the existing behavior for a non-callable `load` (the whole operation entry is dropped, so lookups report "not supported" instead of failing mid-review).

## Test plan

- Added `an operation with a non-callable watch hook is dropped, not wrapped` to `publicApiRobustness.test.ts`, mirroring the existing `load` coverage.
- Confirmed the new test fails on current `main` (TypeError path) and passes with the fix.
- `bun test packages/hunk/src/extensions/runExtension.test.ts` — 51 pass, 0 fail.
- `bun run typecheck`, `oxfmt --check`, `oxlint --deny-warnings` all clean on the changed files.
- Added a patch changeset.

Fixes #763

### DIFF
--- a/.changeset/reject-malformed-vcs-watch-hooks.md
+++ b/.changeset/reject-malformed-vcs-watch-hooks.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Extension VCS operation registration now requires `watchSignature` and `watchPlan` to be absent or callable. A non-callable hook is dropped along with its operation at registration instead of throwing a `TypeError` the first time watch planning invokes it.
--- a/packages/hunk/src/extensions/publicApiRobustness.test.ts
+++ b/packages/hunk/src/extensions/publicApiRobustness.test.ts
@@ -352,6 +352,36 @@ describe("registerVcsAdapter with junk", () => {
     expect(registry.vcsAdapters[0]?.adapter.operations).toEqual({});
   });
 
+  test("an operation with a non-callable watch hook is dropped, not wrapped", () => {
+    const { registry, issues } = loadFactory(
+      (hunk: { registerVcsAdapter: (adapter: unknown) => void }) => {
+        hunk.registerVcsAdapter({
+          id: "hg",
+          name: "Mercurial",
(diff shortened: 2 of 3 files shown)


benvinegar:
Thanks!

PR #1085, opened 2026-09-10
fix(ui): keep the selected file visible in the sidebar during held navigation

### LINKED ISSUE #1084: Sidebar snaps to the top and loses the selected file when `.`/`,` is held down
## What happened

While reviewing a changeset with more files than the sidebar can show, holding down `.` (next file) sometimes scrolls the file list back to its top. The main review stream keeps moving to the right file, but the sidebar no longer shows the selected row. Once it is in that state, further `.` presses do not bring the selected row back; pressing `,` to step back onto a row that is still mounted recovers it.

## What I expected

The sidebar keeps the selected file's row inside its viewport on every step, the same way it does when pressing `.` slowly.

## Reproduction

1. Open a diff with 40+ changed files in a terminal wide enough for the sidebar (160+ columns) and short enough that the list overflows (24 rows).
2. Hold `.` for a second or two.
3. Watch the sidebar: with some probability it jumps back to the first file and the selected row (marked `▌`) disappears from the pane.

A deterministic reproduction is to paste a burst of dots into the terminal, or in a PTY test to write `".".repeat(40)` in one chunk: the sidebar stays at the top and the selected row is never revealed.

## Environment

- Hunk `v0.22.0` (commit `3fdfde2d`)
- macOS, iTerm2 under the `herdr` multiplexer; not terminal-specific
- Installed from source
- Key repeat at the fastest System Settings values (`KeyRepeat=2`, i.e. one repeat every 30 ms; `InitialKeyRepeat=15`). This matters: the renderer paints at 30 fps, so a 30 ms repeat already outpaces frames slightly and the bug shows up often. At the macOS default (`KeyRepeat=6`, 90 ms) it needs the review pane to drop frames first, so it is rarer but still reachable, and a pasted burst of dots reproduces it on any setting.

## Analysis

The bundled sidebar reveals the selected row with `scrollChildIntoView(fileRowId(selectedFileId))` from a `useEffect` (`packages/hunk/src/extensions/default/ui/sidebar/FileSidebars.tsx`). OpenTUI's `scrollChildIntoView` reads `child.y` and `child.height`, which are only synced from Yoga during a frame's `updateFromLayout`; a renderable created since the last frame still has `_y = 0`.

The sidebar is windowed (`buildSidebarRenderWindow`, visible range plus 4 rows of overscan plus the selected row). Key repeat is faster than the renderer's frame interval, so several selection commits land between two frames. Each one scrolls a row and mounts a new overscan row that has no layout yet. Once the selection steps onto such a row, the reveal computes `childTop = viewportTop - scrollTop`, treats the row as being above the viewport, and scrolls by `-scrollTop`, which lands on 0. From then on every newly selected row is a fresh mount with `childTop == viewportTop`, so the reveal computes a delta of 0 and the list stays stuck at the top.

Two related observations while tracing this:

- The sidebar subscribes to `viewport.on("layout-changed")` and `viewport.on("resized")`, but OpenTUI 0.5.6 emits `layout-changed` only on the root renderable and per-node size changes as `resize`. Neither subscription fires, so the render window's viewport height only becomes known after the first scrollbar `change`.
- The documented extension recipe in `docs/extensions.md` uses the same `scrollChildIntoView`-in-an-effect pattern and has the same exposure for windowed lists.

## Proposed fix

Reveal the selected row from the fixed-row geometry the render window already owns (entry index × row height) and `scrollTo` the nearest edge, instead of reading the row's rendered position. Keep the reveal pending until a measured viewport confirms the row is visible so the first paint and a clamped scroll retry once layout settles.

Lawgk:
Fixes #1084.

## Problem

Holding `.` (or `,`) on a review with more files than the sidebar can show sometimes snaps the file list back to its top, and from then on the selected row stays out of view until the selection steps back onto a row that is still mounted. The review stream itself keeps navigating correctly; only the sidebar loses the selected file.

The sidebar revealed the selected row with `scrollChildIntoView(fileRowId(id))` from an effect. OpenTUI only syncs a renderable's `y`/`height` from Yoga during a frame's `updateFromLayout`, so a row the windowed list mounted since the last frame still reports `y = 0`. Held-down navigation delivers key repeats faster than frames, so several selection commits land between two layouts; once the selection reaches such a row, the reveal computes `childTop = viewportTop - scrollTop`, scrolls by `-scrollTop`, and lands on 0. Every later selection is again a fresh mount with `childTop == viewportTop`, so the reveal computes a delta of 0 and the list stays stuck.

## Approach

- Reveal from the fixed-row geometry the render window already owns. `planSidebarRowReveal` in `sidebarRenderWindow.ts` returns the nearest-edge scroll offset from the entry index and `SIDEBAR_ROW_HEIGHT`, so the reveal never reads a row's rendered position.
- Keep the reveal pending until a measured viewport confirms the row is visible. This also makes the first paint reveal a selection that starts outside the initial viewport, and retries a scroll the scrollbox clamped against a content height the next layout will grow.
- Subscribe to the events OpenTUI actually emits. The previous `viewport.on("layout-changed")` / `viewport.on("resized")` subscriptions never fired in 0.5.6 (per-node size changes emit `resize`; `layout-changed` is root-only), so the render window only learned its viewport height after the first scroll. The sidebar now listens to the scrollbar `change`, `viewport` `resize`, and `content` `resize`.
- Fold the flat and tree projections into one entry list computed in `FlexFileSidebar`, since the reveal needs the same entries the rows render. `FlatFileSidebar`/`TreeFileSidebar` had no consumers outside their re-export and are removed.

Non-goals: the extension-facing `scrollChildIntoView` recipe in `docs/extensions.md` has the same exposure for windowed lists in third-party sidebars; that is a documentation/API follow-up, not changed here.

## Why core

This is the bundled sidebar's own follow-selection behavior inside `packages/hunk/src/extensions/default/ui/sidebar`; no extension API changes.

## Tests

- `packages/hunk/src/extensions/default/ui/sidebar/FileSidebars.test.tsx` (new): drives `FlexFileSidebar` through a selection setter with `@opentui/react/test-utils` and changes the selection without rendering a frame in between. Three of the five cases fail on the old code (stuck at top after outrunning laid-out rows, snap back to top when a far row mounts fresh, no reveal of an initial off-screen selection).
- `packages/hunk/src/ui/lib/sidebarRenderWindow.test.ts`: cases for `planSidebarRowReveal`.
- `test/pty/sidebar-follow-integration.test.ts` (new) with the `createManyFileSidebarRepoFixture` harness fixture: writes a burst of 40 `.` presses in one chunk to a live 160×24 Hunk and asserts the sidebar's `▌` row follows to `file-40.ts`, then to `file-50.ts` without returning to the top. Fails on the old code.

## Validation

| Command | Result |
| --- | --- |
| `bun run typecheck` | pass |
| `bun run lint` | pass |
| `bun run format:check` | pass |
| `bun run deps:check` | pass |
| `bun run knip` | 5 pre-existing findings in `packages/term-video` and `packages/hunk-git`, none in touched files |
| `bun test packages/hunk/src/extensions/default/ui/sidebar packages/hunk/src/ui/lib/sidebarRenderWindow.test.ts` | 20 pass |
| `bun test packages/hunk/src/ui/AppHost.extension-sidebar.test.tsx AppHost.sidebar-*.test.tsx AppHost.responsive.test.tsx AppHost.extension-navigation.test.tsx components/ui-components.test.tsx` | 130 pass |
| `bun run test` | 4198 pass, 19 fail; all 19 also fail on `main` in this environment (git prints localized errors, `/var` vs `/private/var` install-VM path checks, jj-colocated checkout makes VCS detection return `jj`, and `AppHost.file-view-modes` soft-reload dialog timeout) |
| `bun run test:integration` | 163 pass, 5 fail; the 5 (`extensions-integration` trust prompts ×2, `lifecycle` SIGHUP/SIGQUIT/SIGPIPE ×3) fail identically on untouched code |
| `bun run test:tty-smoke` | all 10 skip on macOS: the suite needs util-linux `script -f -e -c`, which BSD `script` lacks; relies on Linux CI |

Platforms: macOS (Apple Silicon, iTerm2). Linux and Windows not tested; the change is React/OpenTUI-only with no platform-specific code.

## Reproduction at a real key-repeat rate

Beyond the single-chunk burst the PTY test uses, I drove a live 180×32 Hunk with one `.` every 30 ms (macOS's fastest Key Repeat setting) and every 90 ms (the default), on two diffs, checking whether the sidebar's `▌` row is on screen after 45 presses:

| Diff | 30 ms repeat, old code | 90 ms repeat, old code | 30 ms repeat, this branch |
| --- | --- | --- | --- |
| 60 one-line files | follows (3/3) | follows (3/3) | follows (3/3) |
| 60 files × 400 lines, split view | **stuck at top (2/2)** | follows (2/2) | follows (2/2) |

The heavier diff makes frames slower than the repeat interval, which is the condition the analysis predicts; the trivial diff keeps up at 30 ms and never hits it.

## Visual evidence

Recorded with `skills/hunk-launch-video`, so every terminal frame is the real TUI driven over a PTY: `hunk diff --mode split` on the 60 × 400-line repo at 180×32, github-dark-default, macOS/iTerm2. First half is `main`, second half this branch, with three unhurried `.` presses followed by `.` held at a 30 ms repeat for 45 presses.

The hold is captured press by press rather than as two stills. Grabbing the terminal grid costs ~0.3 ms, so the 30 ms cadence survives capture, and consecutive identical grids collapse into one keyframe. That collapse is itself the measurement: **45 presses painted 16 frames**, about three presses per painted frame, which is the condition the analysis needs.

- Before: the list tracks for a while, then drops the selected row entirely — the review reaches `file-48.ts` while the pane sits at `file-00.ts`, and later presses never bring it back.
- After: the same hold keeps `▌ file-48.ts` on the pane's bottom edge.

The main pane renders only the file header during the hold in both halves. That is ordinary behaviour when navigation outruns the diff body, unchanged by this PR, and the callout marks the files pane as the subject.

https://github.com/user-attachments/assets/0870cc8c-7431-4f79-a947-19d6a362f0e7

### DIFF
--- a/.changeset/steady-sidebar-follow.md
+++ b/.changeset/steady-sidebar-follow.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Keep the selected file visible in the sidebar while holding down next-file or previous-file navigation, instead of snapping the list back to the top.
--- a/packages/hunk/src/extensions/default/ui/sidebar/FileSidebars.test.tsx
+++ b/packages/hunk/src/extensions/default/ui/sidebar/FileSidebars.test.tsx
@@ -1,0 +1,183 @@
+import { describe, expect, test } from "bun:test";
+import { testRender } from "@opentui/react/test-utils";
+import { act, forwardRef, useImperativeHandle, useState } from "react";
+import { createTestDiffFile } from "../../../../../../../test/helpers/diff-helpers";
+import { resolveTheme } from "../../../../ui/themes";
+import { toReadOnlyFileViews } from "../../../events";
+import { FlexFileSidebar } from "./FileSidebars";
+
+const FILE_COUNT = 40;
+const FRAME_ROWS = 8;
+
+const fileId = (index: number) => `file-${String(index).padStart(2, "0")}`;
+
+const files = toReadOnlyFileViews(
+  Array.from({ length: FILE_COUNT }, (_, index) =>
+    createTestDiffFile({
+      id: fileId(index),
+      path: `src/${fileId(index)}.ts`,
+      before: "export const value = 1;\n",
+      after: "export const value = 2;\n",
+    }),
+  ),
+);
+
+interface SidebarSelectionHandle {
+  select(id: string): void;
+}
+
+/** Host the bundled sidebar behind a selection setter so tests drive it like the review does. */
+const SidebarSelectionHarness = forwardRef<SidebarSelectionHandle, { initialFileId: string }>(
+  function SidebarSelectionHarness({ initialFileId }, ref) {
+    const [selectedFileId, setSelectedFileId] = useState(initialFileId);
+    useImperativeHandle(ref, () => ({ select: setSelectedFileId }), []);
+    return (
+      <FlexFileSidebar
+        files={files}
+        selectedFileId={selectedFileId}
+        selectedHunkIndex={0}
+        theme={resolveTheme("github-dark-default", null)}
+        width={30}
+        keybindings={{ matches: () => false, getKeys: () => [] }}
+        actions={{
+          copyText: () => false,
+          selectFile: () => {},
+          selectHunk: () => {},
+          revealLine: () => {},
+          notify: () => {},
+        }}
+      />
+    );
+  },
+);
+
+/** Return the sidebar row carrying the selected-file marker, if one is on screen. */
+function selectedRow(frame: string) {
+  return frame.split("\n").find((line) => line.includes("▌"));
+}
+
+async function renderSidebar(initialFileId: string) {
+  const handle: { current: SidebarSelectionHandle | null } = { current: null };
+  const setup = await testRender(
+    <SidebarSelectionHarness
+      ref={(value) => {
+        handle.current = value;
+      }}
+      initialFileId={initialFileId}
+    />,
+    { width: 36, height: FRAME_ROWS },
+  );
+  await act(async () => {
+    await setup.renderOnce();
+  });
+
+  /** Change the selection without letting a frame lay the new row out first. */
+  const selectWithoutFrame = async (id: string) => {
+    await act(async () => {
+      handle.current?.select(id);
+    });
+  };
+
+  /** Change the selection and let one frame settle, the way an unhurried key press does. */
+  const selectAndSettle = async (id: string) => {
+    await selectWithoutFrame(id);
+    await act(async () => {
+      await setup.renderOnce();
+    });
+  };
+
+  const frame = async () => {
+    await act(async () => {
+      await setup.renderOnce();
+    });
+    return setup.captureCharFrame();
+  };
+
+  const destroy = async () => {
+    await act(async () => {
+      setup.renderer.destroy();
+    });
+  };
+
+  return { selectWithoutFrame, selectAndSettle, frame, destroy };
+}
+
+describe("FlexFileSidebar selected-row reveal", () => {
+  test("follows an unhurried walk down the list", async () => {
+    const sidebar = await renderSidebar(fileId(0));
+    try {
+      for (let index = 1; index <= 12; index += 1) {
+        await sidebar.selectAndSettle(fileId(index));
+      }
+      const frame = await sidebar.frame();
+      expect(selectedRow(frame)).toContain(`${fileId(12)}.ts`);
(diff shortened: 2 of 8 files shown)


elucid:
@Lawgk thanks!

PR #1086, opened 2026-09-10
perf(git): parallelize startup reads

meowhib:
## summary

parallelize independent git startup reads while keeping oversized file detection ahead of patch generation
this reduces startup latency by overlapping git stats, config, source metadata, and untracked file discovery

## verification

five fresh process pty runs with the compiled binary:
- one file: 224 → 197 ms
- 240 files: 417 → 395 ms

validated with typecheck, lint, dependency checks, focused tests, and a compiled pty test

### DIFF
--- a/.changeset/tidy-git-startup.md
+++ b/.changeset/tidy-git-startup.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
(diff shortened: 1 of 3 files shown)


benvinegar:
Nice idea!

PR #1095, opened 2026-09-11
feat(ui): host-owned status line with inline prompts, exposed to extensions

elucid:
## Problem

The bottom status row is a filter-only, host-private surface. `hunk log` has a second, hand-rolled footer with its own hand-rolled line editor. Extensions that need a `less`-style prompt (e.g. `elucid/hunk-less-search`) have to fake one: a permanently open bottom pane costing a row for the whole session, a keyboard mode re-implementing a line editor with `█` standing in for a cursor, and a store token deferring navigation to the pane's `useEffect` because only panes hold `actions`.

## Approach

Make the status row one host-owned surface — the **status line** — that the host's own filter, `hunk log` search, and extensions all drive through the same primitive. Host-internal first, then exposed.

- `packages/hunk/src/ui/statusLine/`: a renderer-free store (insertion-ordered **items** + a FIFO **prompt** queue that settles like the dialog queue), a deterministic, theme-free width/priority **layout**, and one `StatusLine` component that owns the focused `<input>` and the badge click-to-exit.
- `App` mounts it in place of `StatusBar`. The file filter is the first prompt consumer (`prefix: "filter:"`, `onChange: review.setFilter`); the residual `filter=foo` and the notice channel are host items derived from state. `focusArea === "filter"` is now derived from "the host filter prompt is current", so focus and the visible input cannot disagree.
- `LogApp` mounts the same component; `hunk.history.search` opens a `/` prompt. `appendSearch` / `backspaceSearch` / `searchEditing` and the second footer are deleted. The retained query stays visible on the row (`/query`) after a search so `n`/`N` have a visible referent.
- **Extension API 26**: `ctx.statusLine.set/clear` (commands, events, keyboard modes) and `ctx.prompts.line()` (commands). No factory-scope writes; items namespace under the extension, clear on registry replacement/teardown, survive content reloads; prompts cancel on any reload. Attribution follows the dialog policy. Malformed items throw, malformed prompt options reject, a throwing `onChange` warns once.
- The vim-navigation example's `:` line moves from a centered `dialogs.input()` modal to `prompts.line({ prefix: ":" })`, the shape it always wanted.

Non-goals: no history extension surface; no prompt `onKey`/history recall; menu entries remain non-contributable.

## Why core

A prompt-shaped interaction and the row it lives on are terminal chrome every review surface shares; the extension system needed the general capability rather than a feature-specific hook. Two host consumers (filter, log search) prove the primitive before any extension uses it.

## Tests

Unit (colocated): `statusLine/layout.test.ts`, `store.test.ts`, `extensionControls.test.ts`, `StatusLine.test.tsx` (replaces the `StatusBar` cases in `ui-components.test.tsx`), `log/controller.test.ts`.
App-level: new `AppHost.status-line.test.tsx` (items from commands/events/modes, registry-replacement clearing, prompt typing/submit/two-step Escape/queueing/reload cancellation/coexistence with the filter).
PTY: new `hunk log` search case in `log-integration.test.ts`; new prompt + overflow case and the updated vim-navigation case in `extensions-integration.test.ts`. Existing filter PTY coverage (`filter-escape`, `key-routing`, `chrome`) unchanged and passing.

One existing test was adjusted: `AppHost.file-views.test.tsx` pressed Tab, Escape, Tab inside one `act`; the mock input parses `ESC` + `Tab` in one chunk as a single alt-chord, and with a store-driven prompt the keys now actually reach the input (the old state-driven bar never rendered between keys inside one `act`, so the typed text never landed and the test passed by accident). It now presses them in separate acts and asserts each state.

## Commands run

- `bun run typecheck`, `bun run lint`, `bun run deps:check` — clean.
- `bun run test` — 4236 pass, 52 skip; only the Jujutsu "logs unexpected source failures" test fails, reproduced alone on `origin/main`. The watch observer-debounce test passes on main and was a real PR regression, now fixed: the host filter opts into surviving reloads with its value and focus; extension prompts still cancel.
- `bun run test:integration` — 174 pass; only lifecycle SIGHUP/SIGPIPE/SIGQUIT fail, all three reproduced with `bun test test/pty/lifecycle.test.ts` on `origin/main`. Focused `watch`, `filter-escape`, `key-routing`, `chrome`, `log-integration`, and `extensions-integration -t "prompt"` PTY runs all pass, including a new 50-column long-attribution/prefix test proving typed input remains visible beside the badge.
- Latest validation uses Bun 1.4.2 and `TMPDIR=/private/tmp` on macOS to avoid the `/tmp` symlink affecting trust-state path assertions.
- `bun run test:tty-smoke` — skipped on macOS (needs util-linux `script`).
- Real TTY (tmux, isolated `XDG_CONFIG_HOME`): `hunk diff HEAD~3 --no-extensions` — `/` opens the prompt with a cursor, typing narrows live, Escape clears then closes, Tab in/out, `filter=…` residual. `hunk log --no-extensions` — `/` prompt, Enter jumps to the match, `/query` stays on the row, `n` repeats.

Platforms: macOS only.

## Follow-ups

- PR 2: `/` → content search, `n`/`N` → match stepping, bundled `search` extension, bundled composition for commands + line highlighters, `selection.files` on the command context.


elucid:
Force-pushed PR 1: removed the local-only design doc from branch history; restored the host filter’s value/focus across content reloads while extension prompts still cancel; fixed Greptile P1 by truncating long prompt lead-ins before they hide the input. Added store/app/layout and real-PTY regression coverage. Typecheck, lint, dependency checks, and all requested focused PTY suites pass. Full suites only retain main-reproduced Jujutsu source-error and lifecycle signal failures; validation details updated in the PR body.

elucid:
The status-line doc example is now self-contained and typechecks under `bun run check:pack`; the Windows jj 5000 ms timeout is an unrelated runner-speed flake.

PR #1099, opened 2026-09-11
Make daemon/client build skew visible and resolvable with hunk daemon status/restart

elucid:
## Problem

After a Hunk upgrade, a daemon from the previous build keeps running for as long as any old window holds it open. New windows that connect to it are refused at the hello (revision mismatch) or — worse, as in the recent `hunk show` incident — accepted at the hello and then have their registration rejected one parser deeper because #1064 changed a payload shape without bumping `HUNK_SESSION_DAEMON_VERSION`. In both cases nothing told the user what happened, agents got a generic "incompatible or missing required support" message, and the only way to replace the daemon was `kill`.

This PR makes version skew visible on every surface and gives users one supported way to resolve it. **It does not replace the daemon automatically**: old and new windows are peers, and the daemon is replaced only when the user runs `hunk daemon restart` and confirms.

### Guard against the root cause
- **Wire-fixture bump guard.** `wire.snapshot.test.ts` builds the fullest registration + snapshot payload for every launch mode (all three review-descriptor kinds incl. `displayRevision` and comparison commits, catalog, capability digest, experimental features, live comments with `parentId`, review notes, …) through the real `createSessionRegistration` / `updateSessionRegistration`, round-trips it through the daemon parsers, and compares it to `fixtures/session-wire.v<N>.json`. Changing the payload without bumping `HUNK_SESSION_DAEMON_VERSION` fails with instructions; `bun run generate:session-wire` regenerates. Compile-time exhaustive key lists tie the corpus to the session types so a new field cannot be left out. `AGENTS.md` documents the rule.

### Diagnosability
- Wire parsers now record the innermost rejecting parser and key path; the daemon logs `rejected registration from session <id>: parseExtensionReviewDescriptor returned null at info.review` under `HUNK_DEBUG=1` (parser name and path only, never payload contents).
- A post-hello `1008 "Incompatible session registration."` / `"…snapshot."` close is now a warning-bearing close instead of a silent reconnect; the reconnect loop keeps running so a daemon restart recovers the window.

### Revision-tolerant admin scope (`@hunk/session-broker`)
- Two actions, `status` and `stop`, behind the ordinary signed caller handshake and on-disk credentials, authenticated by a **second `SessionBrokerAuthenticator`** whose fixed contract uses `SESSION_BROKER_ADMIN_SCOPE_VERSION` (=1) in place of the app revision, on its own paths (`/session-admin/*`). Its caller sessions are unknown to the main authenticator and cannot reach `/session-api`; the ordinary hello still requires an exact match. Response schemas are frozen per scope version.
- `status` reports daemon revision, app version, pid, uptime, and attached sessions with the revision each producer presented in its hello. `stop` sends the signed ack, then closes producers with `"Session daemon restarting."` and runs the existing graceful shutdown.
- Covered in the Bun/Node runtime conformance fixtures (`admin-scope` mode): a caller from another revision reaches `status`/`stop`, is refused for `list`, and an admin caller session cannot reach the API.

### Direction-aware window notices (TUI)
- On a refused hello the broker client shows the existing generic notice, probes the admin scope, and refines it:
  - **window newer:** `Not connected to the session daemon (daemon build 0.21.1, this window 0.22.0). Run `hunk daemon restart`.` — keeps polling and attaches on its own once the daemon is replaced. When both builds report the same package version, the revision is added (`0.22.0 (revision 14)`), which is exactly the incident's shape.
  - **window older:** `This window is on an older Hunk build than the session daemon. Relaunch it to reconnect (notes in this window will be lost).` — backs off to a 30 s poll.
  - registration rejected after the hello: `The session daemon rejected this window's registration; run `hunk daemon restart`.`
- The notice is **sticky** (new slot in `App`, distinct from the 4 s timed session notice) and clears when the link reaches `connected`. It flows `SessionBrokerClient.subscribeConnectionNotice` → `useHunkSessionBridge` → `App`; UI code never touches the connection. Only genuine faults still reach the console.

### Agent-facing structured error
- Every `hunk session *` command now fails with `DaemonBuildMismatchError` (`kind: "daemon-build-mismatch"`) carrying `daemon`, `cli`, `attachedSessions` (count + list when the admin scope answered), `launch` metadata for a pre-admin daemon, and `recommendedAction: "restart-daemon" | "use-newer-hunk"`. Under `--json` it is returned **in-band on stdout** with exit 1.
- The generated review skill tells agents to report both builds and the attached-window count and **ask** before `hunk daemon restart --yes`, and to re-run `hunk session list` afterwards rather than relaunching windows.

### `hunk daemon status` / `hunk daemon restart`
- `status [--json]`: daemon build, uptime, pid, attached windows (marked `(older build)` when they could not reconnect to a daemon started from this CLI); launch-metadata fallback for a pre-admin daemon; "No session daemon is running." with exit 0.
- `restart [--yes] [--json]`: prints the summary, asks `Restarting will disconnect N attached windows. …Continue? [y/N]` (non-TTY without `--yes` exits 1), **takes the daemon launch lock before stopping anything** so no attached window can respawn the old binary, stops via the admin scope, waits for the port to clear, spawns the replacement from this CLI's own binary, waits for health, releases the lock, reports the new daemon. A pre-admin daemon (the one being upgraded from) is stopped by SIGTERM to the pid in its launch metadata only after a **second explicit confirmation**; nothing ever signals a pid otherwise.

### Notes
- `HUNK_SESSION_DAEMON_VERSION` stays 15: nothing on the producer wire changed. The admin scope is versioned independently.
- The daemon can impersonate another revision through an internal, undocumented `HUNK_INTERNAL_SESSION_DAEMON_VERSION` override so cross-process skew tests can run without building a second binary; the install-VM upgrade fixture regex was updated to recognize the renamed built-revision declaration.
- Bun 1.4's `bun test --no-orphans` kills every descendant of a spawned process the moment it exits, so the restart integration tests run the restart in-process through the same command implementation the CLI calls (the CLI contract itself is covered by spawned help/non-TTY runs).

## Verification
- `bun run typecheck`, `bun run lint`, `bun run deps:check` clean.
- `bun run test`: only pre-existing failures on this macOS machine (install-VM harness path tests, one jj test); everything else green including the new `test/session/daemon-restart.test.ts` end-to-end (old daemon + attached old window + refused new window → restart → new window attaches, old window told to relaunch, pid/revision changed).
- `bun run test:integration`: new `test/pty/daemon-version-skew-integration.test.ts` passes for both directions and for the notice clearing after restart; the remaining PTY failures (lifecycle signals, extension trust prompt) reproduce on `main`.
- `bun run test:session-broker-node` and `bun test ./test/review-conformance` green.
- Manual, from a real shell with a daemon started at revision 14: `hunk daemon status`, `hunk session list` (text and `--json`), `hunk daemon restart` (non-TTY refusal), `hunk daemon restart --yes`, then `hunk daemon status` showing the matched replacement — all as described above.


elucid:
Rebased onto `main` now that #1095 and #1096 are in (both squash-merged). The branch is the previously-reviewed skew work replayed on the new base, with one adaptation: PR #1095 deleted `StatusBar`, so the sticky daemon notice now rides the host status line as its own `host:daemon` item at priority 2 rather than being folded into the timed notice channel — a transient notice must not mask a persistent "this window cannot join the daemon" state. Commit subjects are unchanged; `chore(session): fold the skew wording note into one changeset` is preserved.

`BUILT_SESSION_DAEMON_VERSION` stays 15 — no daemon-forwarded payload changed, and the skew detection here is capability-based.

Local checks on the rebased tree: `bun run typecheck`, `bun run lint`, `bun run deps:check` clean; `packages/hunk/src/session`, `packages/session-broker/src`, and the status-line / interactions / search / watch AppHost suites pass (656). Full validation left to CI.

PR #1102, opened 2026-09-12
fix(pager): render conflicted Git diffs

0xferrous:
## Summary

When Git is paused during a conflicted merge or rebase, `git diff` emits a combined diff using `diff --cc`/`diff --combined` headers and `@@@` hunk syntax. With `core.pager = "hunk pager"`, Hunk previously treated this input as an empty review:

> No files match the current filter.

The pager/difftool configuration was valid; Hunk’s patch sanitizer did not support Git’s combined conflict format.

## Fix

Normalize combined Git patches into the existing unified patch format before parsing:

- Convert combined file and hunk headers to standard `diff --git`/`@@` syntax.
- Project the result against the first parent.
- Omit lines deleted only from other parents.
- Preserve conflict markers and partially resolved changes.
- Handle binary combined conflicts with no unified hunk.
- Handle `--combined-all-paths` by retaining the first parent’s path.
- Leave ordinary unified patches unchanged.

This keeps the behavior in the shared patch-sanitization layer used by Hunk’s existing consumers.

## Reproduction

1. Configure Hunk as Git’s pager:

   ```sh
   git config --global core.pager "hunk pager"
   ```

2. Start a conflicting merge or rebase.
3. Run:

   ```sh
   git diff
   ```

Before this change, Hunk displayed an empty review. After this change, it displays the conflicted file and its conflict markers.

## Validation

- Focused parser and static pager tests: 42 passed
- `bun run typecheck`
- `oxfmt --check`
- `git diff --check`
- Pre-commit formatting and lint hooks passed
- Manual real-PTY reproduction at 120×34 terminal dimensions
- Added a `patch` Changeset for `hunkdiff`

Visual evidence was captured from the real Hunk TUI for the blank before state and rendered after state.

before:
image

after:
image

### DIFF
--- a/.changeset/bright-diffs-conflict.md
+++ b/.changeset/bright-diffs-conflict.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Render Git combined conflict diffs from `git diff` during paused merges and rebases.
--- a/packages/hunk/src/core/patch/gitFormat.test.ts
+++ b/packages/hunk/src/core/patch/gitFormat.test.ts
@@ -301,3 +301,145 @@ describe("sanitizeGitPatchText", () => {
     );
   });
 });
+
+describe("combined Git patch normalization", () => {
+  test("converts a two-parent conflict diff into a first-parent unified diff", () => {
+    const combinedPatch = [
+      "diff --cc notes.txt",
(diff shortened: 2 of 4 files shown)


benvinegar:
Nice. Is this ready? (currently marked as draft)

Issue #1106, opened 2026-09-15
0.22.0 Nix flake package: hunk patch fails with missing hunkdiff sibling binary


schickling-assistant:
## Repro

```sh
nix run github:modem-dev/hunk/v0.22.0 -- patch /tmp/smoke.patch
# error: unable to execute '…/bin/hunkdiff': No such file or directory
```

The flake's `nix/package.nix` installs a single compiled binary as `$out/bin/hunk`, but the `patch` subcommand expects a `hunkdiff` sibling next to the executable. The npm package ships both `hunk` and `hunkdiff` bins (same `./bin/hunk.cjs`, argv[0]-dispatched).

## Regression

v0.21.1 works with the identical install layout (`hunk` + hidden `.hunk-wrapped`, no `hunkdiff`): `hunk patch file.patch` renders the review TUI. v0.22.0 fails as above. `hunk --version` / `--help` / `skill path` still work on 0.22.0.

## Environment

x86_64-linux, Nix (nixpkgs nixos-unstable). Happy to test a fix.

<details>
<summary>Posted on behalf of @schickling</summary>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.udckw4nk |
| `session` | unknown |
| `agent_persona` | generalist |
| `agent_supervisor` | unavailable |
| `agent_tool` | OMP |
| `agent_tool_version` | 18.1.14 |
| `agent_runtime` | OMP 18.1.14 |
| `agent_model` | openai-codex/gpt-5.6-sol |
| `worktree` | 2026-09-12-tmux/2026-09-15-external-flakes |
| `tooling_profile` | dotfiles@183d780 |
</details>

schickling-assistant:
Fix proposed in #1108 (installs the hunkdiff alias in nix/package.nix; verified hunk patch renders from the fixed package on x86_64-linux).


<details>
<summary>Posted on behalf of @schickling</summary>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.udckw4nk |
| `session` | unknown |
| `agent_persona` | generalist |
| `agent_supervisor` | unavailable |
| `agent_tool` | OMP |
| `agent_tool_version` | 18.1.14 |
| `agent_runtime` | OMP 18.1.14 |
| `agent_model` | openai-codex/gpt-5.6-sol |
| `worktree` | 2026-09-12-tmux/2026-09-15-external-flakes |
| `tooling_profile` | dotfiles@183d780 |
</details>

PR #1108, opened 2026-09-16
fix(nix): install hunkdiff alias so hunk patch works

schickling-assistant:
## Problem and user impact

Since 0.22.0, `hunk patch <file>` (and other review flows that respawn the TUI) fail under the Nix flake package with:

```
error: unable to execute '…/bin/hunkdiff': No such file or directory
```

`hunk --version` / `--help` / `skill path` still work. v0.21.1 works with the identical install layout. Reported as #1106.

## Approach

The review flows resolve the TUI through a `hunkdiff` lookup, mirroring the npm package's dual `hunk`/`hunkdiff` bins (same `./bin/hunk.cjs`). The Nix `installPhase` only installed `hunk`, so the lookup fails. This change installs `hunkdiff` as a symlink to the wrapped `hunk` binary (wrapper env, including `HUNK_INSTALL_SOURCE=nix`, is preserved through the symlink).

Non-goals: no behavior change to any other install path (npm, curl, Homebrew); no new package output.

## Tests and manual checks performed

- `nix build .#hunk --no-link` on x86_64-linux: builds clean.
- `hunk patch /tmp/smoke.patch` from the fixed package with a clean `PATH` (package bin only): renders the review TUI, exit 0. Before the fix: the `unable to execute …/bin/hunkdiff` error.
- `hunk --version` and `hunkdiff --version` both report the packaged version.
- Bun unit test suite not run: the change touches only `nix/package.nix` (no TypeScript), and the failure/success is fully determined by the installed file layout, verified above.

## Platforms

- Tested: x86_64-linux (Nix).
- Not tested: aarch64-linux, x86_64-darwin, aarch64-darwin (same derivation shape; relies on CI).

## Follow-up

Once merged, downstream Nix consumers can advance past the v0.21.1 hold.

<details>
<summary>Posted on behalf of @schickling</summary>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.udckw4nk |
| `session` | unknown |
| `agent_persona` | generalist |
| `agent_supervisor` | unavailable |
| `agent_tool` | OMP |
| `agent_tool_version` | 18.1.14 |
| `agent_runtime` | OMP 18.1.14 |
| `agent_model` | openai-codex/gpt-5.6-sol |
| `worktree` | 2026-09-12-tmux/2026-09-15-external-flakes |
| `tooling_profile` | dotfiles@183d780 |
</details>


benvinegar:
Thanks!

PR #1111, opened 2026-09-18
fix(ui): register one terminal blur listener per diff pane

### LINKED ISSUE #1107: Resizing a review can emit MaxListenersExceededWarning and corrupt the display
### What happened?

Resizing a Hunk review can print the following warning over the bottom of the alternate-screen UI:

```text
(node:<pid>) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 blur listeners added to [CliRenderer]. MaxListeners is 10.
```

The warning text becomes interleaved with the UI, and subsequent output is incomplete or incorrectly drawn.

<details>
<summary>GPT Diagnosis</summary>

Each mounted `DiffSectionBody` [registers a `blur` listener on the shared renderer and removes it on unmount](https://github.com/modem-dev/hunk/blob/ee556ac841c1b5a872b80f247816a71e8104605f/packages/hunk/src/ui/diff/DiffSectionBody.tsx#L183-L189). Once 11 file bodies are mounted, this exceeds the EventEmitter default warning threshold of 10. [Node's EventEmitter documentation](https://nodejs.org/api/events.html#eventsdefaultmaxlisteners) notes that this is a warning threshold rather than a hard limit. Since Hunk does remove the listeners on unmount, this may be intentional listener fan-out rather than a retained memory leak.

A width change also [remounts the diff content using a key that includes `width`](https://github.com/modem-dev/hunk/blob/ee556ac841c1b5a872b80f247816a71e8104605f/packages/hunk/src/ui/components/panes/DiffPane.tsx#L2575-L2582), which makes resizing the visible trigger.

The `(node:<pid>) ...` line is the runtime's default warning output to `stderr`. Hunk/OpenTUI renders the alternate screen through `stdout`, while both streams target the same terminal. The warning therefore bypasses OpenTUI's framebuffer, writes at the current cursor position, and moves the cursor without the renderer knowing. Since subsequent frames may update only changed cells, overwritten or stale cells can remain until a full repaint.

OpenTUI also [forwards process warnings to `console.warn`](https://github.com/anomalyco/opentui/blob/v0.5.6/packages/core/src/renderer.ts#L986-L988), but `openConsoleOnError` [opens the console only through the error handler](https://github.com/anomalyco/opentui/blob/v0.5.6/packages/core/src/renderer.ts#L936-L942). It therefore does not prevent the runtime warning from interfering with the active terminal screen.

</details>

### Steps to reproduce

1. Open a review containing enough small changed files for at least 11 file sections to be mounted.
2. Resize the terminal or multiplexer pane horizontally.
3. Observe the warning at the bottom of the UI and incorrect subsequent rendering.

### Expected behavior

Resizing should redraw the review without emitting a listener warning or corrupting the display. Ideally, terminal blur handling would use one renderer-level listener rather than one listener per mounted file body.

### Version

Hunk `0.22.0`, source commit `ee556ac841c1b5a872b80f247816a71e8104605f` (Nix installation).

speedsharmaai:
Fixes #1107

Every mounted `DiffSectionBody` was adding its own `blur` listener on the renderer, so once 11+ files are on screen Node prints the MaxListenersExceededWarning right over the UI.

Only one file can have row hover at a time anyway, and `DiffPane` already clears it through `clearAddNoteHoverForScroll`. So I moved the listener up to the pane and pointed it at that same function (renamed it to `clearAddNoteHover` since it's not only scroll now). Didn't want to just bump `setMaxListeners`, that would only hide it.

Tests:
- 14-file pane registers one blur listener (14 on main)
- blur still hides the `[+]` affordance
- PTY test that opens 14 files and checks the raw output for the warning. I had to read it with `session.subscribe` because the next repaint wipes the warning off the screen. Fails on main, passes here.

typecheck / lint / format / deps:check are clean. A few tests in `bun run test` and `test:integration` fail for me locally (jj, install-vm, extension trust prompt, lifecycle signals) but they fail the same way on main, so I think it's just my setup. Only tried this on macOS.


benvinegar:
Great find, thanks!

PR #1112, opened 2026-09-19
fix(pager): handle plain text output through pipes

0xferrous:
When Hunk wrote a large plain-text document through process.stdout, Bun could report success before all of the data had reached the pipe. Hunk then called process.exit(0), discarding the unwritten remainder. Large output was silently truncated for commands such as Git, LazyGit, and `hunk pager | less`.

Write directly to stdout's file descriptor and keep writing until all bytes are delivered. Treat EPIPE as a normal early exit when a downstream command stops reading, such as `hunk pager | head`.


benvinegar:
Good find, thank you!

0xferrous:
failure seems unrelated

PR #1116, opened 2026-09-21
fix(cli): resolve bundled skills by proximity, not candidate order

shashank-100:
## Problem

`resolveBundledSkillPath` nests its search loops candidate-outer, ancestor-inner, so each candidate *shape* is walked to the filesystem root before the next shape is tried at all:

```ts
for (const root of roots) {
  for (const relativePath of relativeCandidates) {   // candidate is the outer walk
    const resolvedPath = findRelativePathFromAncestors(root, relativePath);
```

With `root = process.execPath` (`~/.local/bin/hunk`), the generic `skills/<name>/SKILL.md` shape is tested at `~/.local/bin`, `~/.local`, `~`, `/home`, `/` — all before `hunkdiff/skills/<name>/SKILL.md` is tried at `~/.local/bin`, where `scripts/install-bin.ts` actually stages the skills for a source install.

**User impact:** anyone with an unrelated `~/skills/hunk-review/SKILL.md` — a personal skills folder, or a Hunk checkout at `~` — runs `bun run install:bin` and then gets that unrelated file from `hunk skill path`, not the one just installed.

The `hunkdiff/` nesting exists precisely to avoid claiming generic skill names beside the executable (see the comment at `scripts/install-bin.ts:46-47`); the resolver's loop order defeated that intent.

## Approach

Invert the nesting so every candidate shape is tested at one ancestor before the walk moves up — proximity wins over candidate order — and order the candidates most-specific-first so a directory holding several layouts resolves to the one that names Hunk explicitly.

`findRelativePathFromAncestors` now takes the candidate list rather than a single path. It has one caller, so no other resolution behavior changes.

**Non-goals:** the set of supported layouts is unchanged, and no discovery root is added or removed. This only fixes which of several existing matches wins.

## Why core

Skill resolution is CLI startup behavior in `core/run`, not renderer or extension surface.

## Tests

Added two regression tests in `packages/hunk/src/core/run/paths.test.ts`:

- a staged skill beside the binary beats a generic `skills/` directory further up (the reported bug);
- the packaged `node_modules/hunkdiff/` layout beats a generic `skills/` sibling at the same ancestor.

Both were confirmed to **fail on the unfixed resolver and pass with the fix** by reverting `paths.ts` to `main` and re-running, so they guard the behavior rather than describe it. This path previously had no coverage for a competing ancestor.

## Commands run

```
bun run typecheck    # clean
bun run lint         # 0 warnings, 0 errors
bun run deps:check   # no violations (467 modules, 1972 dependencies)
bun test packages/hunk/src/core/run/paths.test.ts   # 12 pass, 0 fail
```

The full `bun run test` suite was not run locally; relying on CI for it.

## Platforms

Verified on macOS (darwin 27.0.0). Not run on Linux or Windows; the change is `node:path`/`node:fs` only and adds no platform assumptions, and the new tests build their fixtures with `path.join` rather than literal separators.

## Notes

Supersedes #1007, which was closed in favor of a fresh PR. The candidate ordering requested in review there — `hunkdiff/skills`, then `skills`, then `node_modules/hunkdiff/skills`, with the second test covering it — is incorporated here.


benvinegar:
Thanks!

